Issue #120, opened 2013-03-24
PretzelExpress - tray app which makes user experience better


shiftkey:
Emulate the behaviour of IIS Express, but leverage the existing work of pretzel.
### Why?
- Because scribble needs something more polished than opening a console and showing the raw output.
- Because I'm lazy and want it to "remember" what sites I was testing across sessions
### Features
- single-instance app - subsequent startups should create a new "site" 
- some internal storage for tracking the active sites - Akavache?
- tray icon should behave like IISExpress (list active sites, add site, stop site)

PR #156, opened 2013-08-05
[WIP] Pretzel Express. For all your fast baked goods, you can rely on Pretzel Express*

AwesomeGuyWhoIsAwesomeAtStuff:
#120 requested a "pretzel express", or "pretzel with a gui thats easier to spin up than command lines which are for idiots"

- [x] Stores sites details (directory, port)
- [x] Doesn't use Akavanche like the initial selection because lulzoverkill
- [x] Runs multiple sites real good now
- [x] remember status of window (hidden/visible) on launch
- [ ] Tray icon have more control of stopping/starting sites

I'm really unhappy with this [one line](https://github.com/AwesomeGuyWhoIsAwesomeAtStuff/pretzel/blob/pexpress/src/Pretzel.Logic/Commands/CommandParameters.cs#L16), because without it Pretzel Express can't handle multiple sites nicely, but it may break Pretzel actual.

* Not to be confused with StarTrack Express, Planet Express, Pony Express or other baked good based expresses.


AwesomeGuyWhoIsAwesomeAtStuff:
Tried, failed, thats OK.

PR #160, opened 2013-10-24
Implemented Code highlighting in the Liquid engine

LodewijkSioen:
Implemented code highlighting using a .net port of Pygments, which is the same engine that is used by Jeckyll.

I had to do some voodoo for ILmerge to work, but it works.


LodewijkSioen:
dawson-crying-dawsons-creek-1445426-640-480-540x4051 1

Seems like I have to upgrade my blog to use Hilight.js then :smile:

laedit:
Not necessarily, we have used highlight.js only because it support the HTML 5 syntax for source code.
But you can create your own liquid block to use pygment, Pretzel now supports [plugins](https://github.com/Code52/pretzel/pull/157) in the `_plugins` folder thanks to @miguelerm.

Issue #173, opened 2014-10-10
create command does not work as expected


kodfodrasz:
Using `pretzel create` the results are not as one would expect from `pretzel --help` output.  

I have noticed the following problems:
 * the --drafts option is ignored
 * the generated default site contents contain several errors for both razor and both liquid engine
   * sitemap.xml is badly formed (both)
   * rss.xml fails to compile (razor only)
 * the --directory does not work as expected (this is also a documentation issue - there is no specification for operation, and jekyll has no --directory option)

PR #178, opened 2014-10-21
Create command fixes

kodfodrasz:
Some minor fixes to the `create` command not working as expected out of the box templates did not transform for default freshly created site: RSS and Sitemap template fixes.
`--drafts` was ignored in `create` althoug listed at the command line help. Now it creates (empty) `drafts` folder

(partially) fixes issue #173 reported by me for these problems.


laedit:
Thanks!
I will look into the third point of #173.

PR #185, opened 2014-12-27
Allow all markdown files to be processed

laedit:
*.md, *.mkd, *.mkdn, *.mdown and ".markdown are now allowed.

### DIFF
--- a/src/Pretzel.Logic/Templating/Context/PathExtensions.cs
+++ b/src/Pretzel.Logic/Templating/Context/PathExtensions.cs
@@ -5,7 +5,7 @@ namespace Pretzel.Logic.Templating.Context
 {
     public static class PathExtensions
     {
-        private static readonly string[] MarkdownFiles = new[] { ".md", ".mdown", ".markdown" };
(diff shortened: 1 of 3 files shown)


src/Pretzel.Logic/Templating/Context/PathExtensions.cs
```diff
@@ -5,7 +5,7 @@ namespace Pretzel.Logic.Templating.Context
 {
     public static class PathExtensions
     {
-        private static readonly string[] MarkdownFiles = new[] { ".md", ".mdown", ".markdown" };
+        private static readonly string[] MarkdownFiles = new[] { ".md", ".mkd", ".mkdn", ".mdown", ".markdown" };
```
shiftkey: :thumbsup: to the change, but this seems like a [very big yak](http://superuser.com/a/285878) to shave :grinning:

laedit: Thanks for the review!
I think that these extensions are the most used, so it is may be not necessary to add the others?
Or "mdwn" only.

shiftkey: Leave it as-is. If someone wants a different extension, they can add it in a PR.

PR #188, opened 2015-01-08
Handle bool in yaml

laedit:
In a YAML frontmatter like the following:
``` text
-------
active: true
-------
```
Recognize that the value of "active" is a boolean and convert it in the datas for DotLiquid so that it is possible to use it in a template like that:
```cs
{% if page.active == true %}
```
Like in Jekyll.

But this will cause a breaking change if anyone have a 
```cs
{% if page.active == 'true' %}
```
in a template.

@shiftkey, @JakeGinnivan what do you think?
Is it a big problem or a release note could be sufficient?

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -1,0 +1,42 @@
+ 
+# In development
+
+## Breaking changes
+- [#189](https://github.com/Code52/pretzel/pull/189): Handle bool in yaml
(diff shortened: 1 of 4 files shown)


shiftkey:
> Is it a big problem or a release note could be sufficient?

Seems specific enough that the problem and solution could be documented in release notes!

PR #189, opened 2015-01-09
Add support for category in page permalinks

dkarzon:
We use categories in our permalinks because reasons. :8ball: 

Here are the docs: http://jekyllrb.com/docs/permalinks/


laedit:
I'm good with the "show only a specific category" but why the limit to 3?
Could you make it more generic?

dkarzon:
Looks like I might have read that doco page wrong but our usage is to only use the first category in the url so the `:categories` parameter doesn't do what we need.

dkarzon:
Also, with making it generic. I tried only loading it based on the number of categories on the page but if there are no categories it would not remove the `category1` placeholder so it would fail when creating the directory.

laedit:
I'll try to provide a generic solution, but if I fail I'll merge your PR.
Is that ok for you?

PR #194, opened 2015-01-24
Fix WebSequenceDiagram style bug:the style was not used

laedit:
Now the default style is really 'default', like indicated on https://www.websequencediagrams.com/embedding.html.
And with "@@sequence style", 'style' will replace 'default'.

### DIFF
--- a/src/Pretzel.Tests/Extensibility/Extensions/WebSequenceDiagramsTests.cs
+++ b/src/Pretzel.Tests/Extensibility/Extensions/WebSequenceDiagramsTests.cs
@@ -19,7 +19,7 @@ b->a: bar
 <p>world</p>";
 
             const string expected = @"<p>hello</p>
-<div class=""wsd"" wsd_style=""""><pre>a->b: foo
+<div class=""wsd"" wsd_style=""default""><pre>a->b: foo
 b->a: bar
 </pre></div>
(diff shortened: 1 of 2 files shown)


src/Pretzel.Tests/Extensibility/Extensions/WebSequenceDiagramsTests.cs
```diff
@@ -58,5 +58,26 @@ public void Multiple_blocks_are_converted_to_multiple_diagrams()
...
+
+            const string expected = @"<p>hello</p>
+<div class=""wsd"" wsd_style=""mscgen""><pre>a->b: foo
+b->a: bar
+</pre></div>
+<p>world</p>
+<script type=""text/javascript"" src=""http://www.websequencediagrams.com/service.js""></script>";
+
+            var markdown = transform.Transform(input);
+            Assert.Equal(expected.RemoveWhiteSpace(), markdown.RemoveWhiteSpace());
```
JakeGinnivan: ApprovalTests is awesome for this sort of test. https://github.com/approvals/ApprovalTests.Net

laedit: Thanks for the tip!

PR #202, opened 2015-02-03
Add ContentExcerpt feature for pages

AndreyAkinshin:


laedit:
Thanks for your PR :smiley: 
Just one question, why ContentExcerpt and not Excerpt like with [Jekyll](http://jekyllrb.com/docs/posts/#post-excerpts)?
Also can you add a test with the use of page/post.content_excerpt to display the excerpt?

AndreyAkinshin:
Ok, I will do it.

laedit:
Excellent, thanks!

PR #203, opened 2015-02-04
Support for 404 pages

dkarzon:
This is more a feature of Github pages than Jekyll itself but adding support for custom 404 pages from the Pretzel web host.
https://help.github.com/articles/custom-404-pages/
If there is a 404.html file in the site it will server that when it can't find the correct content.


laedit:
Nice one.

PR #204, opened 2015-02-04
Add InvaritantCulture for some tests.

AndreyAkinshin:
Some tests are Culture specific. For example, they fail when CurrentCulture is ru-ru.


laedit:
Thanks, that was needed.

PR #206, opened 2015-02-04
Code block support

AndreyAkinshin:
Now code blocks are highlighting with prettify


laedit:
This is awesome, I think this syntax works with both liquid and razor.
But there is a few things:
 - there is one test failing on [AppVeyor](https://ci.appveyor.com/project/laedit/pretzel/build/0.2.1.84#L100). I have improved the build, now if a test fail the all build will fail.
 - I think it's too linked to pretiffy, I think it can be the default choice but we can add the possibility to override the highlighter, like in Jekyll?

For the last one, if you haven't the time I will look into it.
I want to make a very flexible highlighting:
 - Support different highlighter
 - Be able to parse different syntax:
   - {% highlight c# %}
   - {{c#}}
   - ```c#

@shiftkey, @JakeGinnivan, what do you think?

src/Pretzel.Logic/Extensions/YamlExtensions.cs
```diff
@@ -107,7 +109,11 @@ public static string ExcludeHeader(this string text)
             if (m.Count == 0)
                 return text;
 
-            return text.Replace(m[0].Groups[0].Value, "").Trim();
+            text = text.Replace(m[0].Groups[0].Value, "");
+            var lines = text.Split(new[] { "\r\n", "\n" }, StringSplitOptions.None).ToList();
+            while (lines.Any() && string.IsNullOrWhiteSpace(lines.First()))
```
shiftkey: Could this be done in a more LINQ-friendly way?

```
var nonEmptyLine = lines.Where(x => !String.IsNullOrWhiteSpace(x));
return String.Join(Environment.NewLine, nonEmptyLines).TrimEnd();
```

shiftkey:
For reference, this is how Jekyll does it:

```
{% highlight ruby %}
def foo
  puts 'foo'
end
{% endhighlight %}
```

A couple of quick thoughts on the proposed syntax (contrasting against the above sample):

 - I've stopped doing "four spaces to indicate code" in Markdown for a while now - it feels fiddly unless you've set your editor up right. I feel the code fencing is so much nicer for managing this pain (shift-TAB to remove the whitespace, live a happy life) but that's specific to GitHub-Flavoured Markdown (which I don't believe is supported in MarkdownDeep
 - One of the other benefits of the previous point is that you can specify the language inline

> \`\`\`ruby

   That feels so much friendlier than needing to wrapp the language name in `{{ foo }}`
 - I do like dropping the "close tag" syntax

src/Pretzel.Logic/Extensions/YamlExtensions.cs
AndreyAkinshin: You can't remove empty line in the middle of the text. My `while` removes only first empty lines. So, `SkipWhile` is needed. I will fix it.

AndreyAkinshin:
@laedit, @shiftkey, thank you for feedback. I will try to add suport of differenet syntax for declare code blocks. About highlighters: I suggest to use some predefine set of highlighters and choose it in the config like in the following line:

```yml
highlighter: prettyify
```

What kind of additional highlighters is needed?

src/Pretzel.Logic/Extensions/YamlExtensions.cs
shiftkey: Good point, :thumbsup: to `SkipWhile`

AndreyAkinshin: Done (commit 0e5338f30042e8403b41b0fd89fe0d756af29def)

AndreyAkinshin:
There is a problem with the MarkdownDeep library. The [issue](https://github.com/toptensoftware/markdowndeep/issues/19) about support for backtick fenced code blocks have been fixed only at [24 Feb 2014](https://github.com/toptensoftware/markdowndeep/commit/4faf39fe06f49571d20dce65a5c08d8a874f6a0d). But the current [NuGet version](https://www.nuget.org/packages/MarkdownDeep.NET/) of the library (0.5) is obsolete (last updated at August 20, 2012). So, it is impossible to implement nice support of markdown codeblocks with this version.

I suggest switching the project to [CommonMark.NET](https://github.com/Knagis/CommonMark.NET) for markdown processing. Advantages:

* Great compatibility
* Great performance
* Full [CommonMark](spec.commonmark.org/) specification support
* Active developing

What do you think?

AndreyAkinshin:
Also I suggest using of [CommonMark Spec 0.17](http://spec.commonmark.org/0.17/) without specific highlighter. For example, lines 

	```ruby
	def foo(x)
	  return 3
	end
	```

will transform in 

	<pre><code class="language-ruby">def foo(x)
	  return 3
	end
	</code></pre>

Next, a pretzel user can add any highlighter yourself. What do you think?

shiftkey:
I'd totally be down for switching over to a more modern Markdown library

laedit:
That is a great idea!
I'm good with that.

laedit:
It seems that [prettify](https://google-code-prettify.googlecode.com/svn/trunk/README.html) always need the `class="prettyprint"`.
But there is [highlightjs](https://highlightjs.org/usage/) and [prism](http://prismjs.com/#basic-usage) which support the HTML5 convention.

AndreyAkinshin:
> It seems that prettify always need the class="prettyprint".

Yes, but you can add this class with one jQuery line. So, it is not a problem.

laedit:
Good point.
Another way to do it: using an IContentTransform specific for each highlighter, if needed.

AndreyAkinshin:
Done.

laedit:
In that case we can use [highlightjs](https://highlightjs.org/) which does not require JQuery.
Or just indicates to the user how to choose and use an highlighter?

For the plugins it is a bit more complicated :)
We have the PR #157 waiting, but in the meantime you can add a class in the `plugin` folder which will implement `IContentTransform`.
We will transform it to an external plugin (dll) after the acceptation of PR #157.

AndreyAkinshin:
Ok, then I just will change `prettyify`+`jQuery` to `highlightjs`.

AndreyAkinshin:
Done.

AndreyAkinshin:
Any updates?

laedit:
Sorry, I haven't got the time to merge it.
I intend to do it this week end, and to accept the PR #157 to be able to create the PrettifyContentTransformer plugin.

PR #214, opened 2015-02-15
Fixes #212 relating to categories being null from incorrect syntax in posts

### LINKED ISSUE #212: If a post has invalid categories, it won't render
Forgive me, it's been awhile since I've used Github so I'm not sure how all the cool kids link to stuff properly anymore.

Basically if a post has no category, it'll throw a `NullReferenceException` when evaluating the permalink. The "offending" line (for lack of a better term) is this:
```permalink = permalink.Replace(":categories", string.Join("-", page.Categories.ToArray()));```

https://github.com/Code52/pretzel/blob/master/src/Pretzel.Logic/Templating/Context/SiteContextGenerator.cs#L402

Now, I know I could send in a PR but there would be a couple of things to think about:

* A check before hand, `if (page.Categories != null)` solves the issue, easy done, right? Well, does that then necessitate a check for each of the date properties?
* Investigate why `page.Categories` isn't initialised, because in theory it is in the constructor for Page.

vikingcode:


laedit:
Thanks, that was quick!
But i'm not comfortable with the fact to let `:categories` in the url if there no categories because Pretzel screwed up.
I think  `Categories` should always be at least an empty collection.

vikingcode:
Ah, I hadn't thought of that scenario (I don't use categories on my permalinks). Do you think instead of a null check, `if (header.ContainsKey("categories") && header["categories"] is IEnumerable<string>)` would be better?

laedit:
Yes, that would be great.

vikingcode:
Bam, and the issue is gone (I hope!)

laedit:
Excellent, thanks!

Issue #125, opened 2013-04-12
Pretzel cannot serve files with spaces in the name


JakeGinnivan:
If I have a file which is generated with spaces in the name, I get a 404

PR #216, opened 2015-02-15
Support to other default pages

wesleyalcoforado:
- Adds support to other default pages (index.html, index.htm, default.htm) when surfing into a directory.
- Performs a 301 redirect if the url points to a directory without a trailing slash.
- Unescapes Urls with Uri.UnescapeDataString, instead of using replace (bug #125)


laedit:
Thanks! :)

PR #219, opened 2015-02-18
Migrate to new NuGet package restore

khellang:
I've become so spoiled by the beautiful :sparkles: package restore progress dialog in VS that I freaked out when the build seemed to take ages with no notice when restoring NuGet packages.

This moves NuGet package restore to the new "automatic" mode :smile:


khellang:
Oh, and this probably will break the build, since you now need to restore NuGet packages before a regular `msbuild Pretzel.sln`. Simply add

khellang:
Maybe not... :stuck_out_tongue:

laedit:
Thanks for the PR!
I wans't aware of that, you juste have to delete Nuget.Targets and all his references to do that?
It doesn't have impact on build process (I mean in general since the AppVeyor build is fine for the PR)?

khellang:
The only impact is that you have to explicitly restore packages before build (on the build machine, VS does it automatically) using `nuget restore`. Don't know why the build didn't fail. Maybe the packages aren't fully wiped between builds?

khellang:
You can read more about Automatic Package Restore [here](https://docs.nuget.org/consume/package-restore) :smile:

laedit:
Exact, I have enabled the AppVeyor cache for the packages.
Thanks for the link :smiley: 
but Pretzel must be build on different OS with different tools and I don't know if SharpDevelop or MonoDevelop support nuget restore out of the box...

khellang:
That's a good point. I don't think it should matter, as long as they're able to run `nuget restore`, or `mono nuget restore`. Have you tested that XBuild works with the old way?

khellang:
Ah, found [this](http://docs.nuget.org/Consume/Package-Restore/MsBuild-Integrated#mono)

khellang:
Looks like the new restore should work on Mono as well: https://nuget.codeplex.com/workitem/3596

laedit:
Sounds good to me, but I prefer ask the opinion of our xbuild specialist: @vikingcode what do you think?

@khellang can you add `nuget restore` to appveyor.yml file in the `before build`section (like [that](https://github.com/laedit/vika/blob/master/appveyor.yml#L22) in order to avoid any problem if the cache is reset?

vikingcode:
How *dare* you insinuate I'm an expert, let alone an expert on mono! That branch is my entire experience with mono, and half the time I spent mutually complaining to @shiftkey about how much Linux hates us.

I would rather the actual developer experience be magical and pleasant, than hold it back because some extra loops might be needed to build it on an OS that won't be used to develop it. If there has to be a "get_the_linux_build_going.sh", so be it. It already needs messing around to work with various permissions

laedit:
@vikingcode your are the only one that have touched pretzel with mono, so you **are** an expert :smiley: 
@khellang I will merge the branch once you have added the `nuget restore` in appveyor.yml.

khellang:
I'm on it! :+1:

khellang:
Done :sparkles:

laedit:
Sorry for the bad indication, it should be `nuget restore src`.

vikingcode:
Although already merged in, I can confirm `mono .nuget/NuGet.exe restore` works from the src directory under Ubuntu.

laedit:
Thanks for the info :smiley:

PR #221, opened 2015-02-19
New command, 'Hungry' 

vikingcode:
This is literally the recipe that inspired the name and was used for the photo on http://code52.org/pretzel/

### DIFF
--- a/src/Pretzel/Commands/HungryCommand.cs
+++ b/src/Pretzel/Commands/HungryCommand.cs
@@ -59,14 +59,6 @@ Eat the pretzels warm, or reheat them in an oven or microwave.";
         public void WriteHelp(TextWriter writer)
         {
             writer.Write(recipe);
-            return;
-            var textWriter = new StringWriter();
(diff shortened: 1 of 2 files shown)


JakeGinnivan:
Haha, awesome :shipit:

laedit:
Excellent :smiley:

src/Pretzel/Commands/HungryCommand.cs
```diff
@@ -0,0 +1,65 @@
...
+Eat the pretzels warm, or reheat them in an oven or microwave.";
+
+        public void Execute(IEnumerable<string> arguments)
+        {
+            Tracing.Info(recipe);
+        }
+
+        public void WriteHelp(TextWriter writer)
+        {
+            writer.Write(recipe);
```
laedit: This is a little bit annoying when one try to understand how to use pretzel and show help...

vikingcode: I made it less obnoxious by describing when you should be using the command, ala parameters.

laedit:
Perfect!

Issue #192, opened 2015-01-20
Support Sass for CSS generation


toburger:
Jekyll allows besides Less the use of [Sass](http://sass-lang.com/) to generate CSS files.
It would be nice if pretzel supports it likewise.

laedit:
I am good with the idea but I don't know SASS or any of the .net implementation.
Do you used any sass library in .net?
Otherwise I think to use [libsass-net](https://github.com/darrenkopp/libsass-net/).

toburger:
I think the libsass-net wrapper ist the best solution on .NET so far.
I don't know of any purely native .NET lib which is able to generate SASS/SCSS at the moment.

[BundleTransformer.SassAndScss](http://www.nuget.org/packages/BundleTransformer.SassAndScss) is using IronRuby to [compile](http://bundletransformer.codeplex.com/SourceControl/latest#BundleTransformer.SassAndScss/Compilers/SassAndScssCompiler.cs
) SASS/SCSS to CSS, which is in a way "native". But I don't know if it's worth taking a dependency on IronRuby?

JakeGinnivan:
My understanding is a lot of effort is being put into libsass which is a c implementation of the ruby version.
We should use the .net wrapper around the c library imo. Which I think its libgit.net right?

Sent from my Windows Phone
________________________________

shiftkey:
I'd look at https://www.nuget.org/packages/libsassnet - seems a reasonable standalone solution.

shiftkey:
*internet high five for @darrenkopp goes here*

darrenkopp:
:thumbsup: :100:

vikingcode:
This is actually fairly easy to do with libsassnet. It's primarily copy&paste the LessTransform.cs and put in libsassnet calls. 

The issue is with ILMerge and the non-IL binary (LibSass.x86.dll) - it can't embed it and have a funtime with the current build script. There may be ways around that. 

I know Fody/Costura work pretty well for those things, which I've switched to but having other issues around MEF.

darrenkopp:
So, here's what's going to happen to libsass soon, and I think it solves your concerns (and those in the other commit, but not 100% sure).

Currently the way libsass works is a plain ol' clr assembly (libsassnet.dll) which references a managed c++ assembly (libsass.*.dll). I've been working with the libsass team to have official windows dll's for libsass which I will start using (currently libsassnet is even 1 version behind because of a managed c++ limitation with C classes).

[...]

PR #222, opened 2015-02-21
This sassy commit is all the rage this summer.

vikingcode:
Solves #192 - introduces Sass for CSS generation. Uses Libsassnet to provide the Sassyness.

Switches out ILMerge for Fody/Costura. When handling unmanaged assemblies, this is the best solution. This required some modifications to how MEF was used.
It uses a very specific version of Fody/Costura because of issues with VS15, I think. Please don't update it.

### DIFF
--- a/src/Pretzel.Logic/Minification/SassTransform.cs
+++ b/src/Pretzel.Logic/Minification/SassTransform.cs
@@ -1,0 +1,88 @@
+﻿using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.IO;
+using System.IO.Abstractions;
+using System.Linq;
+using HtmlAgilityPack;
+using LibSassNet;
+using Pretzel.Logic.Extensibility;
+using Pretzel.Logic.Templating.Context;
+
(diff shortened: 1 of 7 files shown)


vikingcode:
This should probably also remove ilmerge.exe

vikingcode:
This is also obviously not going to work with mono, since its a native binary for the Windahs Operatung Sostem. Is there any entirely MSIL version of Sass?

kodfodrasz:
Even if it does not work on mono, it should be implemented that if this feature is not used on mono, other features are not affected.

IMHO it would also be nice if the user would be informed about this fact in an informative message instead of a cryptic stacktrace.

(I'm not an experienced DllImport user, so I don't know if this is true for the current state)

laedit:
Thanks for the PR, it looks good.
There are just some changes to do:
 - remove all Fody/Costura, we will stop to merge pretzel (see #266)
 - is it possible to add some tests?

vikingcode:
Just saw this. Nope nope and away!

Issue #215, opened 2015-02-15
Pretzel on Mono


vikingcode:
While a Windows dev myself, I'm a cheap bastard and having a Linux host is the way to go for me. I currently use DigitalOcean, with Jekyll building on git commit/push.

I thought I'd start a bit of a discussion and discovery to whats needed to get Pretzel working on mono.

### So far, to get it building

- Mono doesn't have nuget support? Or nuget is built to a not nuget profile, I'm not sure, removing the nuget task from the csproj makes it build as all the references were there as I'd previously built on .NET. [Apparently nuget does have mono support](https://docs.nuget.org/consume/nuget-faq), so perhaps the nuget.exe in the repo is out of date?

- It doesn't have to be removed, but the ILMerging does not work. This can be left to 'error' out or removed from the csproj. [IL-Repack](https://github.com/gluck/il-repack) or [Fody.Costura](https://github.com/Fody/Costura) might be options.

- Because the ILMerging doesn't work, the MEF stuff doesn't work.
Specifically,   `var first = new AssemblyCatalog(Assembly.GetExecutingAssembly());` would need to be changed to   

		AggregateCatalog first = new AggregateCatalog();
		first.Catalogs.Add(new AssemblyCatalog(Assembly.GetExecutingAssembly()));
		first.Catalogs.Add(new AssemblyCatalog(Assembly.GetAssembly(typeof (SanityCheck))));

Ideally some sort of ILMerging that works with mono would be best. 

### Todo
- [ ] Confirm working on *Nix
- [ ] ILMerging/equiv for single exe
- [ ] Nuget all the things.

vikingcode:
Its actually `NuGet.Targets`, in this case, that was the issue. `LogStandardErrorAsError` was throwing an error, removed that from the Exec statements in NuGet.Targets, and it works again.

vikingcode:
Trying to find the root cause, but mono doesn't seem to be picking up `@(MainAssembly)`, which makes ILRepack fail. 

This means instead of  

     /out:bin\Debug\Pretzel.exe "obj\x86\Debug\Pretzel.exe" <list of assemblies to merge>

It gets

    /out: "obj\x86\Debug\Pretzel.exe"   <list of assemblies to merge>

vikingcode:
Changing it from `@(MainAssembly)` to `$(OutputPath)$(AssemblyName).$(OutputType)` with ILRepack is not overly elegant, but works with both mono and .net

vikingcode:
[...]

PR #229, opened 2015-03-29
Add Scriptcs plugins support

### LINKED ISSUE #217: Plugins/scripts
More of a discussion than anything.

One of the advantages Jekyll (not on GH-Pages) that is inherent from Ruby is the ability to have 'loose file' (ie non-compiled) plugins. Plugins could be a wide range of utility and complexity. Two that spring to mind would be a PNGCrush plugin to minify all your images on compile, and for me I'd like to extend Liquid tags to have a `[video id=xyz]` or similar to easily embed a youtube video.

Food for thought

* IronRuby *might* make it possible to use Jekyll plugins as is (should work on mono too)
* ScriptCS/Roslyn might give the ability to do the same for C# ([should work on mono](https://github.com/scriptcs/scriptcs/wiki/Installing-on-Mac-and-Linux))
* [ClearScript](https://clearscript.codeplex.com/) could do the same for JS.

I think even just a single target (where I think C# would be the more obvious choice) of extensibility would be good.

How would the plugins get called in? Much like `_posts`, it'd be a special, per-site folder, `_plugins` following the Jekyll syntax.

laedit:
Fix #217.

Thanks to @vikingcode  and @glennblock this PR allow Pretzel to have plugins as easily as to create a *.csx file in the _plugins folder.

But like @vikingcode has already said for some parts:
 - the projects are now upgraded to .NET 4.5
 - the total size of all dlls / exes is now about 13 mo (we can gain some size if we make a release for Windows (Roslyin) and another for Linux (Mono.Cecil) but I don't know if it worth it)
 - the merging is no longer supported, ILMerge and [ILRepack](https://github.com/gluck/il-repack) explodes over a StackOverflowException. I haven't tried [Fody.Costura](https://github.com/Fody/Costura) but as stated on #215 it doesn't seems to work for Mono, so between that and the total size of files, I think we should just embrace all the little dlls and stop the merging.

@shiftkey, @JakeGinnivan your thoughts?


glennblock:
Why not have it be an opt-in add on package like Pretzel.ScriptCs. Is that
not possible?

laedit:
Like always you give great advice, I will try that.

glennblock:
@laedit not always, but at least some of the time :-)

It would seem based on what I know about this, that it should be possible to keep the scriptcs dependency out of Pretzel. Maybe allow passing in a catalog, which means just the MEF dependency is sufficient, or have a non-MEF specific plugin interface.

src/Pretzel/Program.cs
```diff
@@ -102,6 +103,7 @@ private void LoadPlugins(string[] commandArgs)
                 if (Directory.Exists(pluginsPath))
                 {
                     catalog.Catalogs.Add(new DirectoryCatalog(pluginsPath));
+                    catalog.Catalogs.Add(new ScriptCsCatalog(pluginsPath, new ScriptCsCatalogOptions { References = new[] { typeof(DotLiquid.Tag), typeof(Pretzel.Logic.Extensibility.ITag) } }));
```
glennblock: Instead of hardcoding the dependency on scriptcs, you could find the catalog and dynamically load it. There's a host of diff strategies you could use to do this. One could be checking if the assembly exists, and if it does you dynamically load it and instantiate yourself via reflection. You could move `ScriptcsCatalogOptions` to a separate `Pretzel.ScriptCs.Contracts` package so that Pretzel and the Pretzel.ScriptCs package can reference it, this way Pretzel can still construct and pass in the options easily.

glennblock:
I comment in the PR on one way you might handle this.

glennblock:
comment[ed]

laedit:
Since we already use MEF both for internals and plugins I was thinking to use it again, like loading the `Pretzel.ScriptCs` if exists and include it in the main catalog. Don't know if the `DirectoryCatalog` with a search pattern like`Pretzel.ScriptCs.dll` will do the trick or if it's better to check and load it manually.
The `Pretzel.ScriptCs.Contracts` is a great idea, I will keep it in mind thanks.

JakeGinnivan:
That would actually be quite cool, grab the ScriptCs plugin then ScriptCs extensibility just works.

glennblock:
Yep!

laedit:
I made a quick and dirty test,  but I face the same problem:
 - if I merge all dlls including `Pretzel.ScriptCs.Contracts` there is a cast exception caused by the class used to transmit parameters to the `ScriptCsCatalog`: the runtime cannot convert the type from the merged `Pretzel` assembly to the original `Pretzel.ScriptCs.Contracts` assembly
 - if I try to not merge `Pretzel.ScriptCs.Contracts`, ILMerge throw an exception:
`Unresolved assembly reference not allowed: Pretzel.ScriptCs.Contracts.`. ILRepack goes to a stack overflow...

Did I miss something?
Or we abandon the merging?

glennblock:
I wasn't suggesting Contracts is merged. Pretzel.ScriptCs.Contracts should
be in a package that Pretzel and Pretzel.ScriptCs both reference. I don't
see how that can work with ILMerge.

laedit:
That's what I was trying to do in my second point: Pretzel.ScriptCs.Contracts isn't part of the Pretzel merge, but int that case ILMerge can't work.
Or at least I haven't figured out how to make it work.

glennblock:
I know very little about ILMerge but can't be much help her.

If an ILMerged assembly must have all dependencies merged, then it sounds
like you won't be able to split out the catalog functionality.

JakeGinnivan:
@simoncropp @distantcam any ideas?

Sent from my Windows Phone
________________________________

glennblock:
Think about this more I have an idea of how you can do this which won't require Pretzel to reference the shared contracts directly. A new assembly will contain a ScriptCsCatalogFactory (the answer to all things!). The factory will manufacture a ScriptCsCatalog and return it casted as a ComposablePartCatalog. If Pretzel needs to pass in types and such it can do it via the factory. The factory only exposes primitive  / framework types, so it can easily be invoked via reflection. Pretzel will dynamically look for the factory and load it and then invoke it to get the catalog. 

The code will look something like this...

```csharp
public class ScriptCsCatalogFactory {
  public static ComposablePartCatalog Create(string[] scriptArgs, Type[] references)
  {
    var options = new ScriptCsCatalogOptions {...};
    return new ScriptCsCatalog(options);
}
```

distantcam:
ScriptCs had issues whenever I tried folding it into another assembly. There's a few ScriptCs classes that need to be reworked, and then there was a showstopper in Roslyn that meant I couldn't continue. Having said that I was trying this on the version of ScriptCs that used the old version of Roslyn. I haven't tried it since the update.

glennblock:
Scriptcs still relies on the old version of Roslyn on Windows. You can
install the ScriptCs.Rosyln.Engines module to try the newer Roslyn.

laedit:
Praise to @GlennBlock and shame on me. Following your comment I was so focused on the idea of using a `Pretzel.ScriptCs.Contracts` assembly to pass parameters that I haven't realized that it was the problem and wasn't really needed. Sorry for the fuss.

Just for testing purpose I tried to merge the `Pretzel.ScriptCs`  and all his references (ScriptCs + Engine.Roslyn) but ILMerge still goes to a stack overflow.

laedit:
Thanks to the recommendations of @ferventcoder, Pretzel will install itself in "C:\tools\Pretzel" (BinRoot or soon ToolsRoot in Chocolatey world) and so does Pretzel.ScriptCs.
@JakeGinnivan, @shiftkey are you ok with that?

JakeGinnivan:
No objections from me

PR #231, opened 2015-03-31
Update WebHost.cs

markvantilburg:
Unused code


laedit:
Instead of commenting the code, can you delete it?

markvantilburg:
created a new patch

PR #233, opened 2015-04-01
Better support for exclude config setting

dkarzon:
Allows the `exclude` config setting to be used as well as `pretzel > exclude` and adds support for folder excludes instead of only by full path.


laedit:
Thanks for the PR!

Good point about the config setting, but instead of duplicating the setting I think we should only accept the same as Jekyll: `exclude`.
And please add or modify tests, we don't accept code that is not tested.

One more thing, is it possible that you update the `include` setting as well?

Thanks.

dkarzon:
That sounds like a better option, I'll update and write some tests for the new structure and update the PR.

dkarzon:
@laedit Not sure why my other commit hasn't come through to the PR but this has been done.

laedit:
Great!
Is it possible to include a folder like you made possible to exclude one?

laedit:
Excellent, thanks!

Issue #247, opened 2015-06-23
Wrong handling of categories


thoemmi:
In 0fca5ab @laedit implemented that a post's categories come from both the folder and the front matter. Jekyll behanves different, and I think Pretzel's implementation is wrong.

In Jekyll's [documentation for variables](http://jekyllrb.com/docs/variables/#page-variables) it says for `page.categories`

> The list of categories to which this post belongs. Categories are derived from the directory structure above the `_posts` directory. For example, a post at `/work/code/_posts/2008-12-24-closures.md` would have this field set to `['work', 'code']`. These can also be specified in the [YAML Front Matter](http://jekyllrb.com/docs/frontmatter/).

This links to the [Front Matter documentation](http://jekyllrb.com/docs/frontmatter/), where it states regarding `categories`:

> Instead of placing posts inside of folders, you can specify one or more categories that the post belongs to. When the site is generated the post will act as though it had been set with these categories normally. Categories (plural key) can be specified as a [YAML list](http://en.wikipedia.org/wiki/YAML#Lists) or a space-separated string.

As far as I understand the docs, categories derive from the directory **only** if there are no categories specified in the front matter.

laedit:
I'll do some tests with jekyll to confirm that and accept your PR id it's ok.
In any case thanks for your implication and PR! :smiley:

laedit:
Sorry the the long silence treatment, I was out of town for a few weeks.
I just tested it and Jekyll use both directory and front matter categories when they are specified.
I'm gonna close the PR but again thanks for your implication.

thoemmi:
@laedit I accepted your answer and continued to use my own patched version. However, today I compared again the output of Jekyll and Pretzel.

In fact, Pretzel works like Jekyll when no paginating is used: both only use categories specified in the frontmatter. However, things are different when using pagination. I created a repro at https://github.com/thoemmi/pretzel-categories-test, alls posts are stored in `_posts\archive\2015`.
Jekyll produces correct categories:
![image](https://cloud.githubusercontent.com/assets/73690/11146527/7e0c1922-8a0f-11e5-8a97-540319fd4e52.png)

But Pretzel adds `archive` and `2015`:
![image](https://cloud.githubusercontent.com/assets/73690/11146536/8e57fe0e-8a0f-11e5-9f6e-d9a4d2b7ffc4.png)

[...]

PR #248, opened 2015-06-28
category handling compliant with jekyll

thoemmi:
Only if the page's front matter doesn't specify categories, extract them from the path. See #247 for reasoning.


laedit:
I just tested it and Jekyll use both directory and front matter categories when they are specified.

PR #249, opened 2015-07-01
Fix for page.url when paginating

thoemmi:
When paginating, currently each page gets a `PageContext` instance differing only in the `Paginator` and `OutputPath` properties. Especially all `PageContext`s share the same `Bag` dictionary.

However, if you want to use `page.url` in your template, you always get the url of the **first** page, even on the successive pages. This makes it impossible to include a [canonical url](https://support.google.com/webmasters/answer/139066?hl=en) in the head of your HTML page.

This PR changes the copy c'tor of `PageContext` so that it receives a **clone** of the original `PageContext.Bag`, and when paginating each `PageContext`'s bag will have its correct url property set.


laedit:
Sorry for the silence, I was out of town for a few time.
Thanks a lot for your PR! :smiley:

PR #250, opened 2015-07-01
Inconsistent page.date handling

thoemmi:
Currently the handling of `page.date` is inconsistent:

1. If there's no date specified in a page's front-matter but is extracted from the filename, `page.Bag["date"]` is a `DateTime`
2. If there's a date specified in a page's front-matter, `page.Bag["date"]` is a `string` (as parsed by YamlDotNet)

When you're using `{{ page.date | date_to_xmlschema }}` on your template, and `page.Bag["date"]` is a `string`, it will be rendered as `Liquid error: Object of type 'System.String' cannot be converted to type 'System.DateTime'.`

This PR fixes this by always adding the date back to `page.Bag["date"]` as `DateTime` (in fact I only had to delete some code in `SiteContextGenerator.cs`. :wink:) Two other tests relied on strings, which I fixed.


laedit:
Excellent, thanks a lot!

PR #260, opened 2015-07-30
Refactor logging

kodfodrasz:
- Add some type safety instead of magic strings,
 - flush after receiving error message to prevent log data loss
   on crashing on uncaught exceptions


kodfodrasz:
Overall using a proper logging library, like Serilog, with ConsoleLogger output would be a better choice, IMHO.

laedit:
Thanks for your PR, that was needed :smiley: 

You are right about the use of a proper logging library, can you create an issue about that so we can start to debate about the library to use?
I have heard only good thing about Serilog but I haven't tested it yet.

PR #263, opened 2015-08-17
Enhanced front matter

### LINKED ISSUE #262: Lists of complex types are ignored in the front matter
If for example I have this list in the front matter:
~~~
---
nestedlist:
 - [this, is, a]
 - [nested, list]
---
~~~
or
~~~
---
adictlist:
 - nr: 1
   name: John Doe
   skip: false
 - nr: 2
   name: John Smith
   skip: true
 - nr: 3
   name: Lady Lorem
   skip: false
---
~~~
then `nestedlist` and `adictlist` will be empty when accessed from the Model.

I have a proposed quick-fix.

kodfodrasz:
Fixes #262.

Adds a few tests to FrontMatter parsing of more complex datastructures, including failing ones for lists of lists and lists of dictionaries.
Adds (an admittedly uggly) quick-fix for handling of lists in the front-matter.

### DIFF
--- a/src/Pretzel.Logic/Extensions/YamlExtensions.cs
+++ b/src/Pretzel.Logic/Extensions/YamlExtensions.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.Linq;
 using System.IO;
 using System.Text.RegularExpressions;
 using YamlDotNet.RepresentationModel;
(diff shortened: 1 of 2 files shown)


src/Pretzel.Logic/Extensions/YamlExtensions.cs
```diff
@@ -73,16 +74,24 @@ private static object GetValue(YamlNode value)
...
-                var listResults = new List<string>();
-                foreach (var entry in list.Children)
-                {
-                    var node = entry as YamlScalarNode;
-                    if (node != null)
+                if (list.Children.All(_ => _ is YamlScalarNode)) {
+                    var listString = new List<string>();
+                    foreach (var entry in list.Children)
+                    {
+                        var node = entry as YamlScalarNode;
```
laedit: is there a way to avoid the double cast (is / as YamlScalarNode)?

laedit:
Thanks for your PR!
It doesn't look so bad for a quickfix :)

src/Pretzel.Logic/Extensions/YamlExtensions.cs
kodfodrasz: There is: If I'd project the items with `as` and filter nulls, and compare the length of the lists. This is more readble in my opinion.

laedit:
Good for me.

PR #269, opened 2015-10-25
Update Rss.liquid

markvantilburg:
small z does not render the timezone.


laedit:
Good catch, tanks a lot! :smiley:

PR #271, opened 2015-10-28
Added missing next and previous items in the page Hash in the Liquid engine

k94ll13nn3:
While browsing the files generated by `pretzel create` I noticed that, after adding a new post, there were no previous and next link at the bottom of the page while the layout contained them.

This small change made those links appeared in the demo site from `pretzel create` when multiple posts are present.

I don't know if it is the right place for the change, and I didn't make tests since the Razor counterpart, which already worked, does not seem to have any.

### DIFF
--- a/src/Pretzel.Logic/Templating/Jekyll/LiquidEngine.cs
+++ b/src/Pretzel.Logic/Templating/Jekyll/LiquidEngine.cs
@@ -62,6 +62,9 @@ namespace Pretzel.Logic.Templating.Jekyll
                 y.Add("title", Context.Title);
             }
 
+            y.Add("previous", pageContext.Previous);
+            y.Add("next", pageContext.Next);
+
             var x = Hash.FromAnonymousObject(new
             {
                 site = contextDrop.ToHash(),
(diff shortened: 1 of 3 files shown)


laedit:
Thanks a lot, that was bothering me for a while but I haven't the time to fix it.
Could you just add one or two tests?

k94ll13nn3:
I added some tests, I hope they are enough and fit the project guidelines. 

I had a question: when I am on the most recent post, I am expecting to have a "previous" link but here it is a "next" link since the next page the page older than the current page. Is it an expected behavior ?

laedit:
Thanks.
It's surprising, but I don't know if it's on purpose.

@JakeGinnivan, @shiftkey do you know if it's the expected behavior?

laedit:
So without response, I propose to change the current behavior:
@k94ll13nn3  since you are already on the subject, could you inverse the content of `GetNext` and `GetPrevious methods in `JekyllENgineBase`?
This way we will have a more coherent result.

k94ll13nn3:
Done.

laedit:
You are awesome, thanks!

PR #273, opened 2015-11-05
Fix post_url tag

### LINKED ISSUE #245: post_url tag issues
Using pretzel 0.2.1.169 (build 175) which I downloaded from [Appveyor](https://ci.appveyor.com/project/laedit/pretzel/build/artifacts) (non script cs zip), the post_url tag doesn't respect the permalink setting and also I am pretty sure it is supposed to generate with a leading slash.

Steps to reproduce

1. `pretzel create`
2. edit _config.yaml to add `permalink: /blog/:year/:month/:title/` at the end (note trailing slash to avoid issue #244)
3. edit _posts\2015-05-05-myfirstpost.md to add `{% post_url 2015-05-05-myfirstpost %}` before hello world
4. `pretzel taste`
5. Click the "My First Post" title to go to http://localhost:8080/blog/2015/05/myfirstpost/

You'll see "2015/05/05/myfirstpost.html" at the top of the post.  This is generating a url using the standard permalink scheme instead of the configured one. Also I am pretty sure that there should be a leading slash.  That is based on how I was always able to use post_url in my jekyll site for github.  Also, the [docs](http://jekyllrb.com/docs/templates/#post-url) show an example of `[Name of Link]({% post_url 2010-07-21-name-of-post %})` which would only make sense if the url had a leading slash.  So, the link should be "/blog/2015/05/myfirstpost/"

laedit:
Fix #245.
Add a TagFactoryBase class which provide SiteContext for Tags which need
it.
The custom Dotliquid is needed until the pull request dotliquid/dotliquid#150 is accepted.


k94ll13nn3:
Is it wanted to have the TagFactoryBase.CreateTag method internal (which prevents tags in plugins to use the TagFactoryBase) ?

laedit:
Thanks a lot for the catch! That was a slip.
I've fixed it and added some tests.

PR #274, opened 2015-11-15
added option only_frontmatter_categories

thoemmi:
As described in #247 I added an option `only_frontmatter_categories` to `_config.yml`. If that option is set to `true`, `post.Categories` only contains categories found in its frontmatter, but none extracted from the post's path.


laedit:
Thanks! :smiley:

PR #277, opened 2015-11-21
allow System.IO.Abstractions in ScriptCs

thoemmi:
I tried to use `IFileSystem` in ScriptCs, but it couldn't find the type.

I added `#r System.IO.Abstractions` to the top of my script, but the script engine tried to load that assembly from the GAC instead of the local folder (I assume the engine has some special handling of files starting with `System.`).

Therefore I added `IFileSystem` to the preloaded types. With this change you can use `IFileSystem` in your ScriptCs plugins.


laedit:
Thanks a lot for your implication, you are amazing! :smiley: 
Does this fix the #278 or it is another issue?

thoemmi:
Thanks for merging :grinning: 
Nope, #278 is unrelated.

Issue #278, opened 2015-11-21
`NullReferenceException` in ScriptCs


thoemmi:
I already mentioned this in #275, but my assumptions were wrong. Therefore this new issue.

The `NullReferenceException` happens in `ScriptCsCatalogFactory.CreateScriptCsCatalog`, i.e. before the script is even executed:
![image](https://cloud.githubusercontent.com/assets/73690/11314691/1d5e0890-8fe8-11e5-81fe-92e6fa7d0f9c.png)

The stacktrace is:
```
   at Roslyn.Compilers.CSharp.CodeGen.CodeGenerator.InitializerDimensions(BoundArrayInitialization init)
   at Roslyn.Compilers.CSharp.CodeGen.CodeGenerator.EmitAllElementInitializers(ArrayTypeSymbol arrayType, BoundArrayInitialization inits, Boolean skipConstants)
   at Roslyn.Compilers.CSharp.CodeGen.CodeGenerator.EmitVectorInitializers(ArrayTypeSymbol arrayType, BoundArrayInitialization inits)
   at Roslyn.Compilers.CSharp.CodeGen.CodeGenerator.EmitArrayCreationExpression(BoundArrayCreation expression, Boolean used)
   at Roslyn.Compilers.CSharp.CodeGen.CodeGenerator.EmitExpression(BoundExpression expression, Boolean used)
   at Roslyn.Compilers.CSharp.CodeGen.CodeGenerator.EmitArgument(BoundExpression argument, RefKind refKind)
   at Roslyn.Compilers.CSharp.CodeGen.CodeGenerator.EmitArguments(ReadOnlyArray`1 arguments, ReadOnlyArray`1 parameters)
   at Roslyn.Compilers.CSharp.CodeGen.CodeGenerator.EmitObjectCreationExpression(BoundObjectCreationExpression expression, Boolean used)
   at Roslyn.Compilers.CSharp.CodeGen.CodeGenerator.EmitExpression(BoundExpression expression, Boolean used)
   at Roslyn.Compilers.CSharp.CodeGen.CodeGenerator.EmitAssignmentValue(BoundAssignmentOperator assignmentOperator)
   at Roslyn.Compilers.CSharp.CodeGen.CodeGenerator.EmitAssignmentExpression(BoundAssignmentOperator assignmentOperator, Boolean used)
   at Roslyn.Compilers.CSharp.CodeGen.CodeGenerator.EmitExpression(BoundExpression expression, Boolean used)
[...]

PR #279, opened 2015-11-22
Removed all configurations from solution except AnyCPU

thoemmi:
The solution still contains configurations for *x86* and *Mixed Platforms*. This PR removes all but *AnyCPU*. See #278 too.


laedit:
Thanks!

PR #284, opened 2015-12-05
New command : create new post

k94ll13nn3:
This is an first draft of the command. This is not yet finished and I have some questions regarding the implementation.

First, in the current state of the commands in Pretzel, if I use the `pretzel.exe newpost "A new post in Pretzel"` syntax for the command, it prevent me from using the `parameters.Path` variable because it will contains the title (because of the [SetPath](https://github.com/Code52/pretzel/blob/5da2f8aa9013ecfca06ebd84145aea587178ca41/src/Pretzel.Logic/Commands/BaseParameters.cs#L54) method). This means that the `-s` option will not be usable too. I could create a new option instead, but in my opinion, in this case, the default argument will be the title and not the path. So I'm not sure what is the best way to do it.

Second, I added all the command logic the the SpiceCommand.cs file but it is not ideal for testing. So I was wondering if I should put the work in a file in the Pretzel.Logic project (in which folder?) and call this class from the command.

Finally, I tried to find a name that continue the cooking theme but I'm not sure that I came up with the best name :)

### DIFF
--- a/src/Pretzel/Commands/IngredientCommand.cs
+++ b/src/Pretzel/Commands/IngredientCommand.cs
@@ -11,8 +11,8 @@ using Pretzel.Logic.Extensibility.Extensions;
 namespace Pretzel.Commands
 {
     [PartCreationPolicy(CreationPolicy.Shared)]
-    [CommandInfo(CommandName = "spice")]
-    public sealed class SpiceCommand : ICommand
+    [CommandInfo(CommandName = "ingredient")]
+    public sealed class IngredientCommand : ICommand
     {
 #pragma warning disable 649
 
@@ -26,7 +26,7 @@ namespace Pretzel.Commands
 
         public void Execute(IEnumerable<string> arguments)
         {
-            Tracing.Info("spice - create a new post");
+            Tracing.Info("ingredient - create a new post");
 
             parameters.Parse(arguments);
 
@@ -61,7 +61,7 @@ namespace Pretzel.Commands
         public void WriteHelp(TextWriter writer)
         {
             writer.Write("   Create a new post\r\n");
-            parameters.WriteOptions(writer, "draft", "-s");
+            parameters.WriteOptions(writer, "drafts", "-s");
         }
     }
 }
--- a/src/Pretzel/Commands/IngredientCommand.cs
+++ b/src/Pretzel/Commands/IngredientCommand.cs
@@ -61,7 +61,7 @@ namespace Pretzel.Commands
         public void WriteHelp(TextWriter writer)
         {
             writer.Write("   Create a new post\r\n");
-            parameters.WriteOptions(writer, "drafts", "-s");
+            parameters.WriteOptions(writer, "newposttitle", "drafts", "-s");
         }
     }
 }
--- a/src/Pretzel/Commands/IngredientCommand.cs
+++ b/src/Pretzel/Commands/IngredientCommand.cs
@@ -28,8 +28,6 @@ namespace Pretzel.Commands
 
             parameters.Parse(arguments);
 
-            var title = parameters.NewPostTitle;
-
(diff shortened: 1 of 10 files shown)


laedit:
Excellent, thanks :smiley: 

About the arguments, I think that we should be coherent with the existing commands: add a new parameter `-t` - `--title`. And by default the title could be "New post" or something like that.
A little warning, you have added the `draft` option to the command writer but there is only the `drafts` option existing.

You are right about the location of the class, it should be in the Logic project. Could you move it to the Logic project and all necessary interface, in order to not have to just reference a class in Logic but have the command directly?

About the name, I think that `spice` is more related to the plugin, the posts are the base of pretzel, without it no cooking is possible :smile: 
What do you think about `Ingredient`?
To go further, I think that a `new` command should be good, allowing to add posts, draft or page for now, but maybe helping to create new plugin (spice :wink:) in the future. What do you think of that?

k94ll13nn3:
I have added the `-n` - `--newposttitle` parameter. I wanted to use `-t` - `--title` but it was incorrectly catch by the method that display the help (because of the `template` parameter).

I have changed the name to `Ingredient`, moved the class in the `Pretzel.Logic` project and added some tests for the command.

laedit:
Thanks for the modification, but I still have some remarks/questions:
 - you didn't tell me what you think about a `new` command, don't you think it will be more appropriate? (`pretzel new ingredient "my awesome post"`)
 - instead of `-n` / `--newposttitle` I think `--posttitle` would be less confusing. Sadly the `-p` or `-t` are already taken. The commands organization is really bad right now, we need to improve that too.
 - have you tried to move `ICommand` to Pretzel.Logic? I think that it could be interesting to provide that as extensiblity. But it can be done another time.

k94ll13nn3:
- I did not understand your comment like that. I think that a `pretzel new ingredient "my awesome post"` command might be more appropriate, and it will be easier to had new tasks related to add something instead of recreating a command each time (like the new plugin that you mentioned). But it may require some modifications in the parameters handling to avoid a `pretzel new --task ingredient --title "my awesome post"` like command, which I find a bit heavy.
- Yeah, I wanted to use `--posttitle` too, but like you said, `-p` was already taken.
- Again did not understand that you wanted to move the `ICommand` too, I think it shouldn't be hard to do it, but maybe it will be better in a new PR, to keep this one focus on the command. What do you think?

laedit:
For the ICommand part, it's just the will to have all the logic part in Pretzel.Logic, and allow the plugins creator to add new command. But you are right, I'll do that another time.

About the `new` command, I see that as a sub command with a bunch of pre-defined commands. But I haven't verified if it's doable in Pretzel right now and I'm afraid it's not the case.
I wanted to do that in order to not be stuck with weird command like for this PR the `n` parameter which represent the title of the post...
But since doing that has a risk to impact all commands system, I'm ok to go with your current `Pretzel ingredient` command for now, and maybe modify it as `Pretzel new ingredient` later when the commands system will be modifiable.

src/Pretzel/Commands/IngredientCommand.cs
```diff
@@ -0,0 +1,43 @@
...
+
+#pragma warning restore 649
+
+        public void Execute(IEnumerable<string> arguments)
+        {
+            Tracing.Info("ingredient - create a new post");
+
+            parameters.Parse(arguments);
+
+            var title = parameters.NewPostTitle;
```
laedit: `title` is not used, could you delete it?

laedit:
I was about to merge the PR but there is still a modification to do.

k94ll13nn3:
What did I forget to change/remove ?

laedit:
I added a comment about the `title` variable in `IngredientCommand`.

k94ll13nn3:
I changed it in 6718e58, should have added a comment, sorry :flushed:

PR #286, opened 2015-12-10
Injectable SiteContext (fixes #285)

### LINKED ISSUE #285: Accessing config in an ILightweightMarkupEngine implementation
I have written a plugin implementing `ILightweightMarkupEngine`. I'd like to make it configurable with options set in `_config.yml`. However, I don't see a way to access the configuration.

I ask to either add the `SiteContext` (or `SiteContext.Config`) to the container so it gets injectable, or add `SiteContext` as an additional parameter to `ILightweightMarkupEngine.Convert()`

thoemmi:
With this PR `SiteContext` is now provided by the MEF container, so it's injectable into any plugin.

That was the simplest solution I could think of to enable #285 

@laedit What do you think?

### DIFF
--- a/src/Pretzel.Logic/Templating/Context/SiteContext.cs
+++ b/src/Pretzel.Logic/Templating/Context/SiteContext.cs
@@ -1,8 +1,11 @@
 using System;
 using System.Collections.Generic;
+using System.ComponentModel.Composition;
 
 namespace Pretzel.Logic.Templating.Context
 {
+    [Export]
+    [PartCreationPolicy(CreationPolicy.Shared)]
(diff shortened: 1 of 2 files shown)


laedit:
Thanks a lot for the PR, but I'm really not comfortable with the export of SiteContext.
I prefer that each plugin has acces only to the parts they need, not the all SiteContext.
And the SiteContextGenerator lost all his purpose.

If you can wait until sunday, I will propose a PR which export only the configuration, is that ok?

thoemmi:
Ok, I'm fine with it. Looking forward for your solution :wink:

PR #287, opened 2015-12-13
Accessible configuration

### LINKED ISSUE #285: Accessing config in an ILightweightMarkupEngine implementation

laedit:
Fix #285
Configuration is now accessible through MEF as IConfiguration.

### DIFF
--- a/src/Pretzel.Logic/Configuration.cs
+++ b/src/Pretzel.Logic/Configuration.cs
@@ -1,0 +1,84 @@
+﻿using Pretzel.Logic.Extensions;
+using System.Collections.Generic;
+using System.IO.Abstractions;
+
+namespace Pretzel.Logic
+{
+    public interface IConfiguration
+    {
+        object this[string key] { get; }
+
+        bool ContainsKey(string key);
+
+        bool TryGetValue(string key, out object value);
+
+        IDictionary<string, object> ToDictionary();
+    }
+
+    public sealed class Configuration : IConfiguration
+    {
+        private const string ConfigFileName = "_config.yml";
+
+        private IDictionary<string, object> _config;
+        private IFileSystem _fileSystem;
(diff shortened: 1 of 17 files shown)


laedit:
@thoemmi I hadn't the time to test it in a real plugin but this should work.
Could you tell me if it's ok for you?

thoemmi:
@laedit I tried it in my plugin, it works great! :+1: :shipit:

thoemmi:
Well, my other plugin crashes when I don't re-compile it. It uses this code
```
if (!context.Config.TryGetValue("redirect_generate_aspx", out generateAspx)) {
    return false;
}
```
which will throw `System.MissingMethodException`:
> Additional information: Method not found: 'System.Collections.Generic.IDictionary`2<System.String,System.Object> Pretzel.Logic.Templating.Context.SiteContext.get_Config()'.

I don't know if this is a showstopper, because introducing `IConfiguration` may break existing plugins.

laedit:
I know that this is a breaking change, but I prefer to keep the config read-only instead of letting it as a dictionary.
If you are ok with that I will merge the branch this week.

thoemmi:
I'm fine with that, I just wanted to make you aware of the issue.  I'm looking forward to the merge. 

Are you going to publish a new release and NuGet package once you've merged the changes?

laedit:
I plan to make a new release in january, I won't have the time to do it until then and I'm waiting that the chocolatey packages are validated.

Issue #270, opened 2015-10-27
Render a page with a layout in a Pretzel plugin.


k94ll13nn3:
I am currently trying to create a category generator plugin for Pretzel and I am facing some problems. I understood that in order to generate files, I had to create a class that inherits `ITransform`. I am able to recover all the categories in my site with `siteContext.Categories` and create empty files according to each category.

Next to that I have a `category_index.html` file in the `_layouts` folder that will be the layout for all my category pages. Thanks to that layout, i can manually create a category page by having a file with only the following content :
```
---
layout: category_index
category: some_category
---
```

The problem is that I want my plugin to generate and render those files and I don't know how to do it. I tried to use some functions from `DotLiquid.Template` and I took a look at `JekyllEngineBase.cs` in order to understand the process but none of my attempts worked.

Is there a way to render a page with a layout in a Pretzel plugin ?

laedit:
Thanks a lot for your interest in pretzel, I see that you already have a nice collection of Pretzel plugins! :smiley: 

Sadly the `ITransform` is out of the generation process and have neither Liquid or  Razor engines.
I haven't got the time to test but I think you have 2 short term solutions:
 - do it manually with an html template and a replace regex
 - use `Template.FileSystem = new Includes(siteContext.SourceFolder, FileSystem);` followed by
```
  var template = Template.Parse(templateContents);
  var output = template.Render(data);
```
 where `templateContents` is the content of your category page and `data` a `Hash` with your data (here layout and category).

But it could be interesting to have access to the current `ISiteEngine` in the `ITransform`, by example by passing it in the `Transform` method. If you are interested by making a PR we can discuss how to do it and what methods of `ISiteEngine` should be made public.

laedit:
Another solution could be to add some extensibility points to `SiteContextGenerator`, before and after the posts/pages/tags/categories generation.
This would allow to add some pages/posts to the site context.

k94ll13nn3:
I have not been able to make the `Template` solution worked on a global level (I have something working but only for my setup). .

I will try to take a look at the `ISiteEngine` solution when I manage to have some free time.

laedit:
Fixed by #288

PR #288, opened 2015-12-23
[RFC] new extension point `IBeforeProcessingTransform`

thoemmi:
This PR adds a new extension point `IBeforeProcessingTransform`. Its method `IBeforeProcessingTransform.Transform(SiteContext)` is called after all pages and posts are loaded. This enables plugins to add additional information to a page's bag.

E.g. I've written this plugin to support **series**, which replaces the `series`tag in the frontmatter with a list of all posts with the same `series` tag:
```cs
public class SeriesPageTransform : IBeforeProcessingTransform {
    private const string SeriesTag = "series";

    public void Transform(SiteContext context) {
        var allPages = context.Posts.Concat(context.Pages);

        var allSeries = from page in allPages
            let series = GetSeriesFromPost(page)
            where series != null
            group page by series;

        foreach (var series in allSeries) {
            var allPagesForSeries = series.OrderBy(p => p.Date).ToList();
            foreach (var page in series) {
                page.Bag[SeriesTag] = allPagesForSeries;
            }
        }
    }

    private static string GetSeriesFromPost(Page page) {
        object obj;
        return page.Bag.TryGetValue(SeriesTag, out obj) ? obj as string : null;
    }
}
```

I'm unsure about the name `IBeforeProcessingTransform` but I couldn't think of something better. (You know, the [two hard things in CS](http://martinfowler.com/bliki/TwoHardThings.html) :wink:). If you know a better name I'd appreciate it.

BTW, I had to make `Configuration` public so I can use it in my unit tests.

### DIFF
--- a/src/Pretzel.Logic/Configuration.cs
+++ b/src/Pretzel.Logic/Configuration.cs
@@ -1,4 +1,5 @@
-﻿using Pretzel.Logic.Extensions;
+﻿using System;
+using Pretzel.Logic.Extensions;
 using System.Collections.Generic;
 using System.IO.Abstractions;
 
@@ -15,7 +16,7 @@ namespace Pretzel.Logic
         IDictionary<string, object> ToDictionary();
     }
 
-    internal sealed class Configuration : IConfiguration
+    public sealed class Configuration : IConfiguration
     {
         private const string ConfigFileName = "_config.yml";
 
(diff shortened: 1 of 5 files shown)


laedit:
Excellent, this could also fix #270. @k94ll13nn3, can you test it for your issue?

You are absolutely right about the name, I can't find a good one either :smile: 
Let's just roll with this one for now, a documentation on the wiki will do fine.

But I don't want `Configuration` to be public, only `IConfiguration` should be.
Could you reverse the change on `Configuration` and create a quick mock for your unit tests?
Or should we ship a `Pretzel.Logic.TestingHelpers` with an 'official' mock like `System.IO.Abstractions` is doing?

thoemmi:
Yeah, naming is hard :laughing: 

I reverted the changes to `Configuration`, it's internal again.

I don't think it's worth to introduce `Pretzel.Logic.TestingHelpers`.

k94ll13nn3:
Yep, the `IBeforeProcessingTransform` solved my problem :+1:

Issue #291, opened 2015-12-30
Razor/cshtml files on Azure


keesschollaart81:
Hi guys, 

Maybe I'm missing something but the docs are (on this topic) not that comprehensive.
I'm using the Razor templating engine, the example project gives me a index.cshtml. After baking the sources it's still index.cshtml but it contains only html. By default Azure does not return .cshtml files (as html).

I solved it by adding a target to the shim project file:
```
<Move SourceFiles="_site\index.cshtml" DestinationFiles="_site\index.html" />
```
But it doesnt feel right because if this is how it is supose to work, why is'nt it in the example project.

Am I missing something?

laedit:
Thanks for reporting it :smiley:
It's weird, the file is normally copied in the _site folder with the .html extension.
Wich version do you use?
From chocolatey or github?

k94ll13nn3:
The [HtmlExtensions](https://github.com/Code52/pretzel/blob/master/src/Pretzel.Logic/Templating/Context/LinkHelper.cs#L22) is actually missing `.cshtml`, which cause the URL to end with `.cshtml` and not `.html` and since [the URL is used](https://github.com/Code52/pretzel/blob/master/src/Pretzel.Logic/Templating/Context/PageContext.cs#L69) to generate the file name, it is causing the file name to end with `.cshtml`.

laedit:
Fixed with #292

PR #292, opened 2016-01-01
Fix on Razor based pages not being renamed to .html

keesschollaart81:
Fix for #291


laedit:
Good one, thanks! :smiley:

PR #294, opened 2016-01-10
Fix failing build on dev machine configured for NuGet V3

TimMurphy:
Pretzel build script (.\BuildScripts\build.cmd) uses NuGet 2.8
(.\src\.nuget\NuGet.exe). If the development machine only has Visual
Studio 2015 then its probably configured for NuGet V3 and therefore
the build fails.

This commit adds `<packageSources> ... </packageSources>` to
.\src\.nuget\NuGet.Config to remove build machine dependencies.


laedit:
Thanks! :smiley:

Issue #296, opened 2016-02-06
ArgumentException when modifying a page while taste is running


hartez:
While running `pretzel taste` with the Razor engine, modifying a page results in the following exception:

````
System.ArgumentException: An item with the same key has already been added.
   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector)
   at Pretzel.Logic.Templating.Razor.ExtensibleProxy`1..ctor(IEnumerable`1 extensibleMethods)
   at Pretzel.Logic.Templating.Razor.ExtensibleActivator..ctor(IActivator defaultActivator, IEnumerable`1 filters, IEnumerable`1 tags)
   at Pretzel.Logic.Templating.Razor.RazorSiteEngine.RenderTemplate(String content, PageContext pageData)
   at Pretzel.Logic.Templating.JekyllEngineBase.ProcessFile(String outputDirectory, Page page, Page previous, Page next, Boolean skipFileOnError, String relativePath)
Failed to process C:\Users\hartez\Documents\Bitbucket\alumnihome\newsite\_site\about.html: System.ArgumentException: An item with the same key has already been added.
   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector)
   at Pretzel.Logic.Templating.Razor.ExtensibleProxy`1..ctor(IEnumerable`1 extensibleMethods)
   at Pretzel.Logic.Templating.Razor.ExtensibleActivator..ctor(IActivator defaultActivator, IEnumerable`1 filters, IEnumerable`1 tags)
[...]

PR #297, opened 2016-02-06
Prevent adding of duplicate tags during PreProcess() in RazorSiteEngine

hartez:
Fixes issue #296


laedit:
You are wonderful, thanks again :smile:

Issue #147, opened 2013-06-26
Sporadically crashes on "Taste" caused by 'Cannot access a disposed object.'


jhovgaard:
I haven't been able to figure out what causes this to happen. I can easly be recreated by simply sending about 100 requests in short a time (keep pressing F5 ;-)) I've attached the full stacktrace. When restarting "taste" everything works again.

#### Stacktrace
![mnrg2zmfxm_733couknlzfufmfpd43a66lyqfq-2c3g](https://f.cloud.github.com/assets/1388347/709698/416611c6-de6d-11e2-9e6e-584ee55adda8.png)

markvantilburg:
This still happens in the current version. I tried with the debugger attached but visual studio was not able to show what object it tries to access.

markvantilburg:
Call stack according to intellitrace 

>	Pretzel.exe!Firefly.Utils.SocketWrapper.Shutdown(System.Net.Sockets.SocketShutdown how = {unknown})	C#
 	Pretzel.exe!<>c__DisplayClass9.AnonymousMethod()	C#
 	Pretzel.exe!<>c__DisplayClass6.AnonymousMethod()	C#
 	Pretzel.exe!Firefly.Utils.SocketSender.SetStateImmediate()	C#
 	Pretzel.exe!Firefly.Utils.SocketSender.DoSendCompleted()	C#
 	Pretzel.exe!Firefly.Utils.SocketSender.SocketEventCompleted()	C#
 	Pretzel.exe!Firefly.Utils.SocketEventWrapper.SocketEventCompleted(object sender = {unknown}, System.Net.Sockets.SocketAsyncEventArgs e = {unknown})	C#

laedit:
This is coming from the embedded server which is no longer supported...
We need to replace it but I haven't found any equivalent.

markvantilburg:
Would it be easy to replace the server with IIS Express (if the user has that installed?)

markvantilburg:
It cannot be embedded currently but we could suggest the user to install it, and use it if it's installed. 

And for example use a command line parameter to enable the use of iisexpress where taste creates an on disk folder and points iis express to there.

laedit:
I don't think having a dependency on a thing as heavy and complex as IIS Express is a good thing.
Specially since the issue here is not blocking.
[...]

PR #298, opened 2016-02-06
Replace Firefly with Nowin

hartez:
Replaces Firefly with [Nowin](https://github.com/Bobris/Nowin)

Addresses issue #147


laedit:
Excellent, thanks!

PR #303, opened 2016-07-08
Issue with post frontmatter and pagination

thoemmi:
I have some problems when using post's frontmatter and paginating. Additional properties I define in the a post's frontmatter are not available when I use pagination.

E.g. I have this post:
```
---
layout: default
title: 'Some post'
customfield: 'bar'
---
## Hello World!
```
And my index.html look like this
```
---
layout: default
paginate: 1
title: 'A different title'
---
{% for post in paginator.posts %}
  {{ post.title }}-{{ post.customfield }}
{% endfor %}
```
Then the first page is rendered with
```
  Some post-
```
but I expect
```
  Some post-bar
```
I'm always confused by Pretzel's code: on the one hand `PageContext` has several properties, and on the other hand it has a `Bag`, where some of the properties are duplicated.

This PR contains "only" a unit test to reproduce the issue. I don't know how to fix it at the moment.

### DIFF
--- a/src/Pretzel.Tests/Templating/Jekyll/LiquidEngineTests.cs
+++ b/src/Pretzel.Tests/Templating/Jekyll/LiquidEngineTests.cs
@@ -2243,5 +2243,72 @@ categories: [{0}]
                 }
             }
         }
+
+        public class Missing_post_frontmatter_when_paginating : BakingEnvironment<LiquidEngine>
+        {
+            private const string TemplateContents = "<html><head><title>{{ page.title }}</title></head><body>{{ content }}</body></html>";
+            private const string PostContents = @"---
+ layout: default
+ title: 'Post {0}'
+ customfield: '{0}'
+---
+## Hello World!";
+            private const string IndexContents = @"---
+ layout: default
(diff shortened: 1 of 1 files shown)


laedit:
Thanks a lot for the reporting 😃 
I know that the `paginator` is kind of "weird" at best, but I hadn't take the time to investigate it.
Seems the time has come!

laedit:
So finally it is not linked to the `paginator`directly but by the way we populate the liquid's template data in `LiquidEngine`: the `paginator` is passed directly without processing the data of each of his posts.

So right now the `customfield` is accessible through `post.bag["customfield"]`.
An evolution could be to pass the `paginator` in `LiquidEngine.CreatePageData` but to call a `ToHash` method, like the one used on `ContextDrop`, which will return a `Hash` of all properties of `paginator` and call `ToHash` on each `Page`.

About `PageContext`, I think it is because Razor is also supported: it doesn't use `page.customfield` but `@Model.Bag["customfield"]` to access to custom values in page. So the properties are needed to facilitate the access to usual values.

PR #304, opened 2016-07-09
Added DateToStringFilter to the LiquidEngine.

tditiecher:
The DateToStringFilter was already implemented but never registered to the engine.


laedit:
Wow, thanks a lot, I didn't know about that! 😃

PR #306, opened 2016-07-12
Added support for specifying frontmatter defaults in _config.yml

### LINKED ISSUE #305: Defining defaults for layout in _config.yml
When trying to convert my Jekyll website to a Pretzel website I noticed that for Pretzel I had to specify the layout in each post while Jekyll has the concept of scoped defaults in _config.yml  as described here:  https://jekyllrb.com/docs/configuration/#front-matter-defaults .

Before I'll try to add this feature to the code, is there a reason (other than time) for not having implemented this?

tditiecher:
Functionality is based on:
https://jekyllrb.com/docs/configuration/#front-matter-defaults

Resolves #305

Note: I removed adding a fixed date of '2012-01-01' to the dictionary if
not specified in the _config.yml. It was never used.

### DIFF
--- a/src/Pretzel.Logic/Configuration.cs
+++ b/src/Pretzel.Logic/Configuration.cs
@@ -1,5 +1,6 @@
 ﻿using Pretzel.Logic.Extensions;
 using System.Collections.Generic;
+using System.IO;
 using System.IO.Abstractions;
 
 namespace Pretzel.Logic
@@ -13,15 +14,25 @@ namespace Pretzel.Logic
         bool TryGetValue(string key, out object value);
 
         IDictionary<string, object> ToDictionary();
+
+        IDefaultsConfiguration GetDefaults();
     }
+
+
+    public interface IDefaultsConfiguration
+    {
+        IDictionary<string, object> ForScope(string path);
+    }
+
 
     internal sealed class Configuration : IConfiguration
     {
         private const string ConfigFileName = "_config.yml";
+        public const string DefaultPermalink = "date";
 
         private IDictionary<string, object> _config;
-        private IFileSystem _fileSystem;
-        private string _configFilePath;
+        private readonly IFileSystem _fileSystem;
+        private readonly string _configFilePath;
 
         public object this[string key]
         {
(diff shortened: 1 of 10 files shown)


src/Pretzel.Logic/Configuration.cs
```diff
@@ -80,5 +87,62 @@ public bool TryGetValue(string key, out object value)
         {
             return new Dictionary<string, object>(_config);
         }
+
+        public IDefaultsConfiguration GetDefaults()
+        {
+            return new DefaultsConfiguration(_config);
```
laedit: Since it is used with the site configuration, is it really necessary to instantiate a new defaults each time?

src/Pretzel.Logic/Configuration.cs
```diff
@@ -13,15 +14,25 @@ public interface IConfiguration
         bool TryGetValue(string key, out object value);
 
         IDictionary<string, object> ToDictionary();
+
+        IDefaultsConfiguration GetDefaults();
+    }
+
+
+    public interface IDefaultsConfiguration
```
laedit: Could you move it to it's own file please?

src/Pretzel.Logic/Configuration.cs
```diff
@@ -80,5 +87,62 @@ public bool TryGetValue(string key, out object value)
...
         }
+
+        public IDefaultsConfiguration GetDefaults()
+        {
+            return new DefaultsConfiguration(_config);
+        }
     }
+
+
+    internal sealed class DefaultsConfiguration : IDefaultsConfiguration
```
laedit: Could you move it to it's own file with the `IDefaultsConfiguration` please?

laedit:
Thanks a lot, you are awesome! 😃 
I've wanted to do that since months but never find the time to do it.
I just have a couple remarks before merging.

src/Pretzel.Tests/ConfigurationTests.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        [Fact]
+        public void DefaultsForScope_should_fallback_to_value_from_empty_path_when_given_path_not_found()
+        {
+            var defaults = _sut.GetDefaults().ForScope("_nonexisting");
+
+            Assert.Equal("default-author", defaults["author"]);
+        }
+    }
```
laedit: Could you add a test which check that values in page/post's front matter have priority against defaults?

laedit:
Excellent, thanks!

PR #309, opened 2016-08-24
Update packages.config

markvantilburg:
remove firefly as it's not used anymore


laedit:
Thanks, I was planning to do that but haven't found the time :smiley:

PR #324, opened 2018-07-01
Switch to new project format

biohazard999:
This PR is converting the project from the old csproj style to the new csproj style.

1. Converted all packages.config references to <PackageReference>
2. Converted to package format
3. Deleted files that were excluded in the old csproj anyway
4. Changed the project to use the latest C# compiler
5. Fixed some warnings that occurred after the project switch

This is the first step of bringing pretzel over to dotnet core. :)


laedit:
Thanks for your interest!
I think the appveyor file needs to be modified to use the Visual Studio 2017 image.

About .net core, sadly there are some parts of Pretzel which use native code, so that will be a bit more difficult.

biohazard999:
Yeah, missed the appveyor part. Will update the PR later this day.

No doubt that that will be harder. But going to netstandard 2.0 is the first step :)

PR #325, opened 2018-07-25
Dotliquid sorting

rolandbaer:
Updated DotLiquid to current version to have changed sorting feature and added test for the new sorting feature


laedit:
Thanks a lot!

rolandbaer:
Thank you for maintaining this software and merging so fast! 👍

PR #327, opened 2019-09-05
Change to the new csproj format & updated nuget packages to prepare for dotnet core

biohazard999:
To move pretzel to dotnet core, this is the first step.

* Switched to new package format
* Updated nuget packages to the latest version (except for System.IO.Abstractions)

The reason to leave out IO.Abstractions is cause unit tests are starting to fail.


laedit:
Thanks! 😃 
By default Appveyor uses the image for Visual Studio 2015, can you update the `appveyor.yml` with `image: Visual Studio 2017` (or 2019) on top to use a more recent Visual Studio?

biohazard999:
I've never used Appveyor before, so this change should do the trick?

laedit:
Perfect, thanks!

Issue #320, opened 2017-05-01
Run Pretzel under .Net Core?


AnonJr:
This is more an alternative to #81 - instead of trying to support running on Mono, maybe go for .NET Core instead? You'll get the same ultimate goal (cross-platform) and you can keep it to a single code base (instead of having a .NET version and a Mono version).

Just a thought for an old, loved project that seems to have had some recent work. :)

laedit:
I would love that, but it will need a lot of changes.
That does not mean that will never happened, but not before a long time 😞

AnonJr:
I understand how that goes. I'll keep following along. :)

biohazard999:
Do you mind if i fork it and push it over to core and release as a global tool?

laedit:
I would love to see that! 😄 
It has been in my work list for so long but never had the time to do it.

biohazard999:
@laedit My plans are: bringing to dotnet core as a global tool and switch to azure devops.

laedit:
You mean replace Appveyor by Azure devops?
Do you know if it is possible to have an account for a Github organization?
I'd prefer to avoid user-specific account to prevent any future issue to access it.
I am trying to activate Github Actions but since I am not an admin I don't have access to the settings of this repo.

shiftkey:
> I am trying to activate Github Actions but since I am not an admin I don't have access to the settings of this repo.

@laedit apologies for that, I should have done it a long time ago. I've given you admin rights to this repository now, let me know if there's anything else you need.

biohazard999:
@laedit Did not use Github actions so far, but for open source it would be a nice fit!
Any suggestions on Microsoft Composition? Move to dependency incjection from dotnet core hosting model or to Microsoft.Composition (which seams dead). I'm currently trying to port to netstandard, but this seams hard so far. (got it compiling and tests running, but composing the catalog fails cause the lack of InheritedExportAttribute that is heavy used)

biohazard999:
Let's get 🥨 some ❤!

laedit:
@shiftkey wow, I was planning to contact you but I have forgotten that you are omniscient 😄 
Thanks a lot!

laedit:
@biohazard999 I must confess that I haven't peek at the code for a long time, but I think that `InheritedExport` is used mainly for extensibility, so if that is the only major change we can remove it and contact every known plugin creators to inform them of the change.

[...]

PR #328, opened 2019-09-05
Move Pretzel.Logic to multitarget net451/netstandard2.0

biohazard999:
#320 

- This add's support for netstandard20 and net451 for the Pretzel.Logic project.
- Cause MEF is not supported in netstandard20 use System.Composition (MEF2) instead.

This is of course an breaking change for Plugin Authors.


laedit:
Thanks a lot!

It appears that GitVersion isn't included in Visual Studio 2019 images, so can you edit the `appveyor.yml` and add the following before the use of GitVersion?
``` yml
  - choco install gitversion.portable
```

I will try to review the PR during the wee-end.

biohazard999:
> Thanks a lot!
> 
> It appears that GitVersion isn't included in Visual Studio 2019 images, so can you edit the `appveyor.yml` and add the following before the use of GitVersion?
> 
> ```yaml
>   - choco install gitversion.portable
> ```

Seams there is an [issue on GitVersion on VS2019 image](https://help.appveyor.com/discussions/problems/24758-gitversion-5-not-working-on-visual-studio-2019-build) so

```yaml
  - choco install gitversion.portable --version 4.0.0
```
Did the trick.
Now I need to tackle xunit...

> 
> I will try to review the PR during the wee-end.

Thanks! :)

biohazard999:
@laedit Hmmm could not get the build to working. but I'm not sure whats failing...

laedit:
Sorry, I didn't get the time to check it this weekend.

The build is weird, I don't see either why it is failing.
Can you try to comment the first line?
``` ps
$ErrorActionPreference = "Stop"
```
Maybe it will show more info.

.gitignore
```diff
@@ -1,3 +1,4 @@
+tools/
```
laedit: This should not be ignored, there are still some tools used during the build.

BuildScripts/AppVeyor-Build.ps1
```diff
@@ -127,12 +127,12 @@ function ExecuteTests($cover)
...
         cinst opencover.portable -y
         cinst coveralls.io -source https://nuget.org/api/v2/
-        & C:\ProgramData\chocolatey\lib\opencover.portable\tools\OpenCover.Console.exe -register:user -filter:"+[Pretzel.Logic]*" -excludebyattribute:*.ExcludeFromCodeCoverage* -target:"%xunit20%\xunit.console.exe" -targetargs:"""src\Pretzel.Tests\bin\Release\Pretzel.Tests.dll"" -noshadow -appveyor" -output:$artifacts\coverage.xml -returntargetcode
+        & C:\ProgramData\chocolatey\lib\opencover.portable\tools\OpenCover.Console.exe -register:user -filter:"+[Pretzel.Logic]*" -excludebyattribute:*.ExcludeFromCodeCoverage* -target:"$tools\xunit.runner.console\tools\net472\xunit.console.exe" -targetargs:"""src\Pretzel.Tests\bin\Release\Pretzel.Tests.dll"" -noshadow -appveyor" -output:$artifacts\coverage.xml -returntargetcode
         & coveralls.net --opencover $artifacts\coverage.xml
     }
     Else
     {
-        &$tools\xunit\xunit.console.exe "$src\Pretzel.Tests\bin\Release\Pretzel.Tests.dll"
+        &$tools\xunit.runner.console\tools\net472\xunit.console.exe "$src\Pretzel.Tests\bin\Release\Pretzel.Tests.dll"
```
laedit: I like the idea to get this dependency directly from nuget, but there are other things to consider:
 - it is used in more places (like build.cmd, I know it is a mess, it should be addressed sometime)
 - other dependencies who fits like GitVersion should be downloaded the same way

BuildScripts/AppVeyor-Build.ps1
```diff
@@ -174,7 +174,7 @@ function Build()
 
             if ($LastExitCode -ne 0) { throw "Building solution failed" }
 
-            ExecuteTests $true
+            ExecuteTests $false
```
laedit: Tests should be executed on AppVeyor

BuildScripts/AppVeyor-Build.ps1
laedit: Nevermind, since GitVersion is needed only on AppVeyor it could still have his special treatment.

laedit:
@biohazard999 Sorry to bother you but in order to avoid mixing all things, I will create another PR dedicated to the build fix.
Could you remove any modification related to it in this PR?
Thanks.

.gitignore
biohazard999: Cause the other files are already under source (and packages.config is force added) this should not harm if we plan to download the build dependencies from nuget anyway

BuildScripts/AppVeyor-Build.ps1
biohazard999: My focus was getting the build to run, so I didn't check the other places it is used so far.

biohazard999:
> @biohazard999 Sorry to bother you but in order to avoid mixing all things, I will create another PR dedicated to the build fix.
> Could you remove any modification related to it in this PR?
> Thanks.

Sure, I'll split it up in two batches if you like.

laedit:
> Sure, I'll split it up in two batches if you like.

No need to, just keep the multi-targets here and I handle the build with #330

biohazard999:
@laedit $ErrorActionPreference = "Stop" did the trick, the testrunner couldn't find the test assemblies (sure, now with multitargeting they moved to `bin\Release\net462`. Should I revert those changes now anyway?

laedit:
@biohazard999 Yes please, my goal right now is to fix the build through #330, merge it to master and then update this pr with master.

laedit:
Thanks for the merge!
It is still failing on `nuget pack` for Pretzel.Logic, I think it is because of the [multi-targeting](https://github.com/NuGet/Home/issues/7765).
Could you replace it with a `dotnet pack` instead?

biohazard999:
@laedit this should do the trick :)

laedit:
That's nice, thanks a lot! 😃 
I will review it this week-end.

src/Pretzel.Logic/Commands/BaseParameters.cs
```diff
@@ -18,23 +21,30 @@ public sealed class BaseParameters
 
         public bool Safe { get; private set; }
 
-        [Export("SourcePath")]
```
laedit: I'd prefer to avoid modification on public api.
MEF2 doesn't handle the export of string?

src/Pretzel.Logic/Configuration.cs
```diff
@@ -55,7 +59,8 @@ private void EnsureDefaults()
             _defaultsConfiguration = new DefaultsConfiguration(_config);
         }
 
-        internal void ReadFromFile()
+        [OnImportsSatisfied]
+        public void ReadFromFile()
```
laedit: I'd prefer to avoid making this public, is it possible for OnImportSatisfied to work on internal or private methods?

src/Pretzel.Logic/Configuration.cs
```diff
@@ -17,8 +19,9 @@ public interface IConfiguration
         IDefaultsConfiguration Defaults { get; }
     }
 
-
-    internal sealed class Configuration : IConfiguration
+    [Shared]
+    [Export(typeof(IConfiguration))]
+    public sealed class Configuration : IConfiguration
```
laedit: Do exported classes need to be public?

src/Pretzel.Logic/Configuration.cs
biohazard999: After a quick check and the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.composition.onimportssatisfiedattribute?view=dotnet-plat-ext-2.1#remarks) it must be public or internal.

laedit:
Sounds globally good to me, except for two things:
- accessibility changes need to be minimal, do all imported/exported properties/classes needs to be public?
- is there no way to load the ScriptCs plugin? What is the issue here?

src/Pretzel.Logic/Commands/BaseParameters.cs
biohazard999: There is probably a more elegant way 🤷‍♂️

src/Pretzel.Logic/Configuration.cs
biohazard999: Just tested it out, seams like it works. Doc's arn't specific about that. But importing constructors need to be public. Seam's extreme weird to me.

biohazard999:
So importing on fields is not supported -> go to properties. Property import must be public.

As for ScriptCs. Yeah that will be the next tricky part.

biohazard999:
I can revert the public class -> internal class. But property import needs to be public anyway.

src/Pretzel.Logic/Commands/BaseParameters.cs
laedit: If I am following correctly, it is caused by `Configuration` which involves `Path`?
If that's the case, I don't see a better way so good to me.

laedit:
Ok for the public properties, but yes please revert the classes when possible, thanks.

src/Pretzel.Logic/Pretzel.Logic.csproj
```diff
@@ -23,23 +23,27 @@
...
     <PackageReference Include="Microsoft.AspNet.Razor" Version="3.2.7" />
+    <PackageReference Include="Microsoft.Windows.Compatibility" Version="2.1.1" />
     <PackageReference Include="NDesk.Options" Version="0.2.1" />
     <PackageReference Include="RazorEngine" Version="3.10.0" />
     <PackageReference Include="SonarAnalyzer.CSharp" Version="7.16.0.8981">
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
     <PackageReference Include="System.IO.Abstractions" Version="2.0.0.116" />
     <PackageReference Include="YamlDotNet" Version="6.1.2" />
+    <PackageReference Include="Microsoft.Composition" Version="1.0.31" />
```
laedit: Shouldn't it be `System.Composition`?

src/Pretzel.Logic/Commands/BaseParameters.cs
biohazard999: No the reason is that i see no way of registering the instance of `BaseParameters` in the container that is created in `Program.cs`. That's why i added the second ctor that pulls out the program arguments from `Environment.GetCommandLineArgs().Skip(1).ToArray()`.

I just checked for exporting strings & named string exports and it seams to work. So that should work and I could remove the SourcePathProvider.

src/Pretzel.Logic/Pretzel.Logic.csproj
biohazard999: Good catch. Heck it's so confusing cause the naming changed so often.

biohazard999:
> Ok for the public properties, but yes please revert the classes when possible, thanks.

Configuration was the only class that was internal and exported. So that should be fixed

src/Pretzel.Logic/Commands/BaseParameters.cs
biohazard999: Removed the `SourcePathProvider`.

biohazard999: I have just one idea to avoid the `Environment.GetCommandLineArgs().Skip(1).ToArray()`:

- Don't use `SatisfyImports` on Program, use `GetExport` to resolve `Program`
- Now Compose and all Part discovery needs to be static
- Save the base Parameters in `Main` in an static variable
- Export BaseParameters from Program's instance and get the value in the ctor.

That way only a single instance of BaseParameters would be created

laedit: Ok, sorry I was focused on the `Path`, I understand now (at least I think I do).
About `BaseParameters`, it is added as Part only to export the path and filesystem, is it necessary to export the entire class in order to export some properties?

src/Pretzel.Logic/Commands/BaseParameters.cs
```diff
@@ -21,20 +24,27 @@ public sealed class BaseParameters
...
             Options = new OptionSet
                 {
-                    {"help", "Display help mode", p => Help = true},
-                    {"debug", "Enable debugging", p => Debug = true},
+                    { "help", "Display help mode", p => Help = true },
+                    { "debug", "Enable debugging", p => Debug = true },
                     { "safe", "Disable custom plugins", v => Safe = true },
                     { "d|directory=", "[Obsolete, use --source instead] The path to site directory", p => Path = p },
-                    { "s|source=", "The path to the source site (default current directory)", p => Path = p}
+                    { "s|source=", "The path to the source site (default current directory)", p => Path = Path = p }
```
laedit: There is a redundancy of Path affectation.

src/Pretzel.Logic/Configuration.cs
```diff
@@ -55,6 +59,7 @@ private void EnsureDefaults()
             _defaultsConfiguration = new DefaultsConfiguration(_config);
         }
 
+        [OnImportsSatisfied]
         internal void ReadFromFile()
```
laedit: Does `OnImportSatisfied` need to be on internal method?
Otherwise, it could be private to avoid any unwanted call.

biohazard999: As stated earlier from the the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.composition.onimportssatisfiedattribute?view=dotnet-plat-ext-2.1#remarks) it must be public or internal.

> The method that this attribute is applied to must be void, parameterless, and either public or internal.

src/Pretzel.Tests/Templating/Context/SiteContextGeneratorTests.cs
```diff
@@ -13,6 +13,7 @@
 using System.Threading;
 using Pretzel.Logic.Extensibility;
 using Xunit;
+using System.ComponentModel.Composition;
```
laedit: It should be `System.Composition`.

biohazard999: There is no need to export the test classes at all. Fixed.

src/Pretzel.Logic/Commands/BaseParameters.cs
biohazard999: No it's not required (I was really struggeling with the NullReferenceException if the export does not work, so I tried any thought it's required). Fixed.

src/Pretzel.Logic/Configuration.cs
```diff
@@ -38,7 +41,8 @@ internal Configuration()
             EnsureDefaults();
         }
 
-        internal Configuration(IFileSystem fileSystem, string sitePath)
+        [ImportingConstructor]
+        public Configuration(IFileSystem fileSystem, [Import("SourcePath")] string sitePath)
```
biohazard999: The sitePath isn't used at all. Should I remove it as well?

src/Pretzel.Logic/Commands/BaseParameters.cs
biohazard999: Good catch again, fixed.

src/Pretzel.Logic/Configuration.cs
laedit: It is used to calculate the `_configFilePath`.

biohazard999: That's correct. My bad. Before it wasn't imported, but now logic moved from `Program.cs` to `ReadFromFile`'s `OnImportsSatisfied`

src/Pretzel/Program.cs
```diff
@@ -17,27 +17,35 @@ namespace Pretzel
...
     {
         [Import]
-        private CommandCollection Commands { get; set; }
+        public CommandCollection Commands { get; set; }
+
+        [Export]
+        public IFileSystem FileSystem { get; set; } = new FileSystem();
+
+        [Export]
+        public BaseParameters Parameters { get; set; }
```
laedit: I don't think that BaseParameters needs to be exported, or am I missing something?

src/Pretzel/Program.cs
```diff
@@ -113,63 +119,81 @@ public void Compose(BaseParameters parameters)
...
-                    Tracing.Debug("Assembly 'Pretzel.ScriptCs.dll' detected but not loaded.");
+
+                    AddScriptCs(configuration, pluginsPath);
                 }
             }
         }
+
+        private void AddScriptCs(ContainerConfiguration configuration, string pluginsPath)
+        {
+            //TODO: Make ScriptCS Working
```
laedit: Since ScriptCs isn't available in .net standard for now, I think the best is to let it works only in net451 and to throw an exception in netstandard (only if the directory is found), until we replace it by [dotnet script](https://github.com/filipw/dotnet-script) or another solution.

src/Pretzel.Logic/Commands/BaseParameters.cs
laedit: The importing constructor shouldn't be needed anymore since only the Parse method is used in Program.

biohazard999: That's what I was talking about before. Now there are 2 instances. One created by Parse one by MEF. If I restructure the bootstrapping I can workout a way that the SourcePath get's exported by Program. But for now the ImportingCtor need's either to stay, or I do the refactoring.

I've updated a version of Program.cs that avoids the circle/recursion problem.

src/Pretzel/Program.cs
biohazard999: No they don't need to be exported

src/Pretzel/Program.cs
biohazard999: It isn't supported by `scriptcs.componentmodel.composition` So it doesn't work with MEF2 at all. There is a contribution project to ScriptCS for MEF2 but it's empty. 😢

src/Pretzel.Logic/Commands/BaseParameters.cs
laedit: Good to me, thanks!

src/Pretzel.Logic/Commands/BaseParameters.cs
```diff
@@ -18,23 +19,26 @@ public sealed class BaseParameters
...
         public string Path { get; private set; }
 
-        [Export]
         public IFileSystem FileSystem { get; private set; }
 
         public List<string> CommandArgs { get; private set; }
 
         private BaseParameters(string[] arguments, IFileSystem fileSystem)
+        {
+            SetDefaults(arguments, fileSystem);
```
laedit: There is no need for a SetDefaults method anymore, no?

src/Pretzel/Program.cs
laedit: Sorry, I was confused. Could you just throw an exception for now and we will find a replacement in another PR.

src/Pretzel.Logic/Commands/BaseParameters.cs
biohazard999: Yupp, refactored.

src/Pretzel/Program.cs
biohazard999: Done.

laedit: Perfect!

laedit:
So I think we are good here, or am I missing something?
I have started to write a [small doc](https://github.com/Code52/pretzel/wiki/Migrate-to-0.8-%5BWIP%5D) with the public differences (for plugins writers mostly) can you tell me if I have forgotten something?
Thanks!

biohazard999:
> So I think we are good here, or am I missing something?
> I have started to write a [small doc](https://github.com/Code52/pretzel/wiki/Migrate-to-0.8-%5BWIP%5D) with the public differences (for plugins writers mostly) can you tell me if I have forgotten something?
> Thanks!

I think we can overcome this limitation with the convention based model. But for now I think that's it.

You should mention that we moved from `System.ComponentModel.Composition` to `System.Composition` `v1.2.0` for now.

biohazard999:
@laedit Anything there holing the PR back?

laedit:
Sorry, didn't get the time to do anything on github today.

Thanks a lot for your hard work! 😄 

Next tasks before releasing the next version (not necessarily by you :) )
- target net462 #335
- replace ScriptCs #336
- replace other dependencies which doesn't support netstandard #337
- use Cake as build definition #329 
- make a dotnet global tool #339

biohazard999:
No worries, I thought you just forgot to merge it 😁

PR #333, opened 2019-09-13
Topic/fix xunit warnings

### LINKED ISSUE #332: Fix warnings in tests produced by xunit analyzer
After upgrading to the latest version of xunit there are a lot of warning about the assertions.

Should we just fix those warning or should we move to a fluent assertion library like [Shouldly](https://github.com/shouldly/shouldly) to make the assertions more readable?

biohazard999:
fixes #332


src/Pretzel.Tests/Pretzel.Tests.csproj
```diff
@@ -19,6 +19,7 @@
     <PackageReference Include="DotLiquid" Version="2.0.314" />
     <PackageReference Include="NDesk.Options" Version="0.2.1" />
     <PackageReference Include="NSubstitute" Version="4.2.1" />
+    <PackageReference Include="Shouldly" Version="3.0.2" />
```
laedit: Is this really used?

biohazard999: Nope.

laedit:
Thanks!

There is still a xunit warning on appveyor build:
> xUnit1013: Public method 'When' on test class 'SpecificationFor' should be marked as a Fact.

Is it possible to indicate to xunit that this is normal?

biohazard999:
> Thanks!
> 
> There is still a xunit warning on appveyor build:
> 
> > xUnit1013: Public method 'When' on test class 'SpecificationFor' should be marked as a Fact.
> 
> Is it possible to indicate to xunit that this is normal?

I can supress it. Adding the `[Fact]` didn't change it. Seams like a bug in the analyzer

laedit:
Perfect, thanks a lot!

Issue #337, opened 2019-09-17
Replace dependencies whish doesn't support netstandard


laedit:
We should replace there dependencies by other that supports netcore, like:
- [WebMarkupMin.Core](https://www.nuget.org/packages/WebMarkupMin.Core) for minification
- [dotless.core](https://www.nuget.org/packages/dotless.Core/) for dotless
- [RazorLight](https://www.nuget.org/packages/RazorLight/) or directly [Microsoft.AspNetCore.Razor.Runtime](https://www.nuget.org/packages/Microsoft.AspNetCore.Razor.Runtime/#show-github-usage) for razore rendering

biohazard999:
Probably we could go a route for dotnetcore2.2 where we can replace some stuff with aspnet core that is supported with fullframework. 
Will require more work and of course more breaking changes, but looking forward to .net5 I think it's the best option

laedit:
If netcore2.2 allows simplifying things, I say go for it.

biohazard999:
@laedit  What about `Ndesk.Options`? They say on their [website](https://github.com/Latency/NDesk.Options) there is support for netstandard, but they point to MyGet and the feed is not public? Should we go for [System.CommandLine](https://github.com/dotnet/command-line-api/wiki/How-To#Add-a-subcommand)?

laedit:
@biohazard999  Good call!
I will always prefer to go for a System.* package than a third party when possible.

biohazard999:
I did a quick look one on RazorLight and Microsoft.AspNetCore.Razor.Runtime today.
It seams like RazorLight it's the better option (much more similar to the old implementation) but it seams like development is dead there (last package upgrade is 2017) 
So going the long route would require more work, but would benifitial in the future.

laedit:
I agree, we should go for Microsoft.AspNetCore.Razor.Runtime.

biohazard999:
Good news!
I've got a version working with the help of [RazorEngine.NetCore](https://www.nuget.org/packages/RazorEngine.NetCore/) seams to be maintained by [fouadmess](https://github.com/fouadmess/RazorEngine) but I need to submit an PR to add custom directives to the Engine

laedit:
Excellent! 😃

biohazard999:
@laedit do you have a project with razor in use that we can use for testing?

biohazard999:
[...]

PR #341, opened 2019-09-18
Upgrade System.IO.Abstractions to 6.0.32

### LINKED ISSUE #340: Upgrade System.IO.Abstractions to >= 6.0
Currently we stick on `System.IO.Abstractions` `v2.0.0.116` which does not support `netstandard`. 
We should investigate why some unit tests are failing.

#337

biohazard999:
fixes #340 
relates #337 

Currently there is one failing unit test. I'm not sure if there is now a introduced but, or the test is not correct. Maybe System.IO.Abstractions mocking is now closer to the actual System.IO when dealing with relative paths. 

The `foldersToDelete` now contains: `c:\\css\\..\\_site` which is the path of the `OutputDirectory`. When it's get's deleted the `Should_Process_Less_Imports_With_Cleanup` test failes.

@laedit Can you look into that?


src/Pretzel.Logic/Import/BloggerImport.cs
```diff
@@ -102,7 +102,12 @@ private void ImportPost(BloggerPost post)
 
             try
             {
-                fileSystem.File.WriteAllText(Path.Combine(pathToSite, Path.Combine("_posts", fileName)), postContent);
+                var path = Path.Combine(pathToSite, Path.Combine("_posts", fileName));
```
laedit: We should avoid the call to `Path.GetDirectoryName`:
``` cs
var postsPath = Path.Combine(pathToSite, "_posts");
var path = Path.Combine(postsPath, fileName);

if(!fileSystem.Directory.Exists(postsPath))
...
```

src/Pretzel.Logic/Import/WordpressImport.cs
```diff
@@ -66,7 +66,12 @@ private void ImportPost(WordpressPost p)
             var postsFolder = "_posts";
             var fileName = string.Format("{0}-{1}.md", p.Published.ToString("yyyy-MM-dd"), p.PostName.Replace(' ', '-')); //not sure about post name
 
-            fileSystem.File.WriteAllText(Path.Combine(pathToSite, postsFolder, fileName), postContent);
+            var path = Path.Combine(pathToSite, postsFolder, fileName);
```
laedit: Same here, we should avoid the call to Path.GetDirectoryName:
``` cs
var postsPath = Path.Combine(pathToSite, "_posts");
var path = Path.Combine(postsPath, fileName);

if(!fileSystem.Directory.Exists(postsPath))
...
```

src/Pretzel.Tests/Minification/LessTransformTests.cs
```diff
@@ -178,6 +178,7 @@ public void Should_Process_Less_Imports_With_Cleanup()
             Assert.False(fileSystem.File.Exists(@"c:\css\style.less"));
             Assert.True(fileSystem.Directory.Exists(@"c:\css\subfolder"));
             Assert.True(fileSystem.File.Exists(@"c:\css\subfolder\anothersubfolder\anything.less"));
+            Assert.Contains(@"c:\_site", fileSystem.AllDirectories);
```
laedit: I think that this is what is failing the test.
Since html isn't processed and all less files outputs in `c:\css`, `c:\_site` is not created.

biohazard999: But it did before, see line 182. I just used `Contains` to see what's on disk.

src/Pretzel.Logic/Import/WordpressImport.cs
biohazard999: But the directory doe's not exist so it failes now. It's a behavior change. `WriteAllText` failes if the directory does not exist.

biohazard999: But we can optimize the directory name creation.

src/Pretzel.Tests/Minification/LessTransformTests.cs
biohazard999: So either the previous assertion was wrong or we fixed a bug that left over empty directories 👌

src/Pretzel.Logic/Import/WordpressImport.cs
biohazard999: I did get you wrong. Never mind 👍

src/Pretzel.Logic/Import/BloggerImport.cs
biohazard999: Fixed.

src/Pretzel.Logic/Import/WordpressImport.cs
biohazard999: Fixed.

src/Pretzel.Tests/Minification/LessTransformTests.cs
biohazard999: I removed the assertions, cause I also think they are wrong

laedit: Sorry, I have missed that line.
I think that the assertion was wrong and there a bug fixed in `MockFileSystem`, anyway, all good now.

PR #343, opened 2019-09-18
Use NUglify instead of AjaxMin

biohazard999:
relates #337


src/Pretzel.Logic/Pretzel.Logic.csproj
```diff
@@ -40,6 +39,7 @@
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
     <PackageReference Include="System.IO.Abstractions" Version="2.0.0.116" />
+    <PackageReference Include="WebMarkupMin.NUglify" Version="2.6.0" />
```
laedit: Why don't use [`NUglify`](https://www.nuget.org/packages/NUglify/) directly?

src/Pretzel.Logic/Minification/JsMinifier.cs
```diff
@@ -20,13 +21,16 @@ public JsMinifier(IFileSystem fileSystem, IEnumerable<FileInfo> files, string ou
...
             var content = fileSystem.BundleFiles(files);
-            var minified =  minifer.MinifyJavaScript(content, codeSettings);
-
-            fileSystem.File.WriteAllText(outputPath, minified);
+            var minified = Uglify.Js(content);
+            if(minified.HasErrors)
+            {
+                foreach(var error in minified.Errors)
+                {
+                    Console.WriteLine(error.ToString());
```
laedit: Should be `Tracing.Error` or Warning depending on what is in `Errors`.
And should we write the minified code if there are errors?

biohazard999: For Tracing sure.
I don't know, I don't know what NUglify does if there are errors. Probably we should write unminified if there are errors?

src/Pretzel.Logic/Pretzel.Logic.csproj
biohazard999: I think you are right, we don't need `WebMarkupMin` at all

src/Pretzel.Logic/Minification/JsMinifier.cs
laedit: Based on [this](https://github.com/xoofx/NUglify/blob/master/src/NUglify/UglifyError.cs#L133), I wonder if we should treat any error as an exception and interrupting the process...
What do you think?

biohazard999: Is minification only used in production builds? If so I don't think it would be a problem. But if it's used in `taste` wouldn't that blow up the watcher?

src/Pretzel.Logic/Pretzel.Logic.csproj
biohazard999: Fixed.

src/Pretzel.Logic/Minification/JsMinifier.cs
laedit: Well, after some search I think that it is not used at all since it is no referenced by another class and it does not implement `ITransform`...
So, for now, let's say that a reporting through `Tracing.Error` will be enough.

biohazard999: Okay, i'll fix this tomorrow

Issue #146, opened 2013-05-31
Is there a way to specify a different _config.yml when baking?


ghost:
I need some ability to specify "environment" when baking / exporting the pages. For example I might need one that's for live (i.e. _config_live.yml) and one for local (i.e. _config_local.yml) etc. Is there a way to do that? So I would expect something like this when I am calling the bake:

For live
.\Pretzel.exe -t razor -c _config_live.yml -d .\live_site
For local
.\Pretzel.exe -t razor -c _config_local.yml -d .\local_site

laedit:
There is nothing for that right now but that's a good idea.
I don't know if I'll have the time to do it, but a PR is always welcome :smiley:

happycollision:
Adding to this conversation, It would be great to see such a feature mimic the Jekyll way of allowing overrides. For example, `pretzel taste --config=_config.yml,_config.dev.yml` would first load the regular config, then override/add any values with the second config.

This is great, for example, for locally testing a site at `localhost:8080/`, but publishing at `example.com/blog/` as Jekyll does by default by referenceing links like `{{ site.baseurl }}/2016/10/01/a-thought-on-pretzel`. Regular config defines baseurl as `"/blog"` and dev config overrides it as `""`.

If I have time to look at the code and see if it is trivial enough for a guy who is completely unfamiliar with programming for Windows, I'll submit a PR sometime.

laedit:
Thanks a lot for your interest in Pretzel! 😄 
I will gladly review your PR.

dalenicholls:
I've written a bat script to get round this issue and use _config_bake.yml & _config_taste.yml file

```
@echo off

set source=C:\src\
set dest=C:\build\

copy %source%_config_bake.yml %source%_config.yml /Y

Pretzel.exe bake %source% --cleantarget --destination=%dest%
```

PR #347, opened 2019-09-18
Use System.CommandLine.Experimental instead of NDesk.Options

### LINKED ISSUE #346: Use System.CommandLine.Experimental instead of NDesk.Options
needed for #337

biohazard999:
fixes #346

This is a major breaking change.

This PR replaces `NDesk.Options` with Microsoft's new `System.CommandLine.Experimental`.

Cause they are fundamentally different in architecture there are some key gotchas to the whole startup architecture as well as argument handling.

**Motivation**
- be netstandard complient
- use something that is supported by Microsoft
- make it more extensible
- allow plugin autors to add additional arguments to commands with ease
- allow plugin authors to create custom commands

Before all commands shared a `BaseParameters` and `CommandParameters` class as well as the `IHaveCommandLineOptions` for extentions to provide custom arguments.

This changed fundamentally. Every command has it's own command line arguments. It is extensible via the `CommandArgumentsExtentionAttribute` and `IHaveCommandLineArgs` exports and interfaces.

Example for a new command:

```cs
[Export]
[Shared]
[CommandArguments(CommandName = BuiltInCommands.Import)]
public class ImportCommandParameters : PretzelBaseCommandParameters
{
        [ImportingConstructor]
        public ImportCommandParameters(IFileSystem fileSystem) : base(fileSystem) { }

        protected override void WithOptions(List<Option> options)
        {
            base.WithOptions(options);
            options.AddRange(new[]
            {
                new Option(new [] {"--importtype", "-i"}, "The import type")
                {
                    Argument = new Argument<string>()
                },
                new Option(new [] {"--importfile", "-f"}, "Path to import file")
                {
                    Argument = new Argument<string>()
                },
            });
        }

        public string ImportType { get; set; }

        public string ImportFile { get; set; }
}

[Shared]
[CommandInfo(CommandName = BuiltInCommands.Import, CommandDescription = "import posts from external source")]
    class ImportCommand : ICommand
{
  [Import]
  public ImportCommandParameters Parameters { get; set; }
}

```
`CommandInfo.CommandName` must match `CommandArguments.CommandName` to get hooked together.
To reduce risk of failure the `BuiltInCommands` class was introduced.

Every command that need's site context can use `PretzelBaseCommandParameters` as a base for the parameters. For extendable parameters authors can use `BaseParameters` as a base class. For complete control plugin authors can implement `ICommandParameters` and `ICommandParametersExtendable` on their own.

**Plugin authors that need custom arguments for an command**

Author's can extend multiple commands at once by using multiple command names:

```cs
   [Export]
    [Shared]
    [CommandArgumentsExtention(CommandNames = new[] { BuiltInCommands.Bake, BuiltInCommands.Taste })]
    public class VirtualDirectorySupportArguments : IHaveCommandLineArgs
    {
        public void UpdateOptions(IList<Option> options)
        {
            options.Add(new Option(new [] { "-vDir", "--virtualdirectory"}, "Rewrite url's to work inside the specified virtual directory")
            {
                Argument = new Argument<string>()
            });
        }

        public void BindingCompleted()
        {
        }

        public string VirtualDirectory { get; set; }
    }
```
The implementation keeps the same except for they just import their new parameters

```cs
  [Export(typeof(ITransform))]
    public class VirtualDirectorySupport : ITransform
    {
        [Import]
        public VirtualDirectorySupportArguments Arguments { get; set; }

        public void Transform(SiteContext siteContext)
        {
            if (string.IsNullOrEmpty(Arguments.VirtualDirectory)) return;
            /****/
      }
    }
```
## New application startup

We use an separate RootCommand to provide the 3 parameters used in startup (majorly plugin discovery). Those get's injected afterwards into the actual rootcommand/subcommands after MEF kicked in. This way we get the benifit of having help available even if MEF discovery should fail in some sort of way. Attributes are processed before the actual command instance get's created. This way separate commands can't break the whole application if MEF binding will fail (at least for the commands, arguments are a different story).

The Logic from `BaseParameters.SetPath` moved to `Program`. Afterward's the `-s` parameter will be set. This way the value is correct before it hit's the arguments parsing. I think this simplifies the overall design of the system.
The reason for this parameter is: we can start pretzel with `pretzel bake c:\foo\myblog` hence no breaking change for pretzel consumers. There is probably a better way of solving this problem in the `System.CommandLine` library, but couldn't figure out how 🤷‍♂️.

**Architectural questions that need to be discussed**

- [x] Should Commands still be in `Pretzel` and should the parameters life there as well (or should we move them to `Pretzel.Logic`. 
  - [x] If Commands still should be there, should we provide an interface for the parameters that get's exported for plugin authors.
  - [] ~~If not, how should plugin authors get needed parameters? We could for example double bind parameters (just use the same names as in the pretzel build in ones), but that could be error prown cause of the modifications that can be made in `BindingComplete`. Exporting for example the `IBakeCommandParameters` would be easy and intuitive.~~
- [x] IConfiguration got `ReadFromFile` method public, this way we get rid of the [Import("SourcePath")] and get a lot nicer lifecycle. It's now 100% clear when configuration get's read. So no weird side effects if `IConfiguration` get's read to early in the MEF lifecycle.
- [x] `IHaveCommandLineArgs` has a weird name. Probably we should rename it to `ICommandArgumentsExtention` that way it would match the attribute `CommandArgumentsExtentionAttribute`
- [x] Cause the new library supports async execution, I've changed the return type to `Task` but i think it would be befinitial to change it to `Task<int>` to support error codes, if we change it anyway.
- [x] I've obsoleted the `Parameters.Path` in favor of the `Parameters.Source` and `Parameters.Destination` properties, cause I think it's much clearer what path we are talking about. It's now also consistent with the command line parameter name.
- [x] I'm not very happy with the inconsistent naming with arguments and parameters so we should be consistent here, cause when we break anyway it should be consistent.

**Things that need to be done before merging**

- [x] Get test converage up
- [x] Make sure we don't loose functionality
- [x] Make old unit tests work
  - [x] Check if we don't miss any important tests
  - [x] Test parameters for extentions
  - [x] Test PretzelCommandHandler
  - [x] Test CommandCollection
  - [x] Test arguments hook in `Program.cs`
  - [x] IntegrationTest composition
- [ ] Document breaking changes

@laedit as always, would be cool if you got an early eye on this. I think this layes out a good foundation for the future of 🥨 ❤

This is an opener for #146


laedit:
Thanks a lot for your hard work!
I am sorry but I will probably don't have the time to check it before this weekend...

biohazard999:
> Thanks a lot for your hard work!
> I am sorry but I will probably don't have the time to check it before this weekend...

No worries, I got plenty work to do as you see my checklist 👷

biohazard999:
@laedit i think I'm ready for review now, probably one test for `Programs.Run` method would be benifitial, but other that that I don't see anything missing on my side.

laedit:
Noted, I should have the time to review it tomorrow :)

laedit:
Sorry, didn't get much time yesterday.
I will try to start the review today, but first your questions:
 - All things needed for extensibility should be in Pretzel.logic, so commands and their parameters too
 - 👍 to rename `IHaveCommandLineArgs` in `ICommandArgumentsExtention`
 - 👍 to support error codes through `Task<int>` in `Program`
 - 👍 to replace `Parameters.Path` by `Parameters.Source` and `Parameters.Destination`
 - 👍 for consistency naming in arguments and parameters

src/Pretzel.Logic/Commands/PretzelBaseCommandParameters.cs
```diff
@@ -0,0 +1,83 @@
...
+            });
+        }
+
+        // Default Option that get's injected from Program
+        public string Source { get; set; }
+        // Default Option that get's injected from Program
+        public bool Debug { get; set; }
+        // Default Option that get's injected from Program
+        public bool Safe { get; set; }
+        [Obsolete("Use '" + nameof(Source) + "' instead.")]
```
laedit: Since we are breaking everything in the next version I think the property can be removed entirely, if it is documented.

src/Pretzel.Logic/Commands/BaseParameters.cs
```diff
@@ -1,67 +1,37 @@
...
+            {
+                factory.CreateExport().Value.UpdateOptions(Options);
+            }
         }
 
-        public static BaseParameters Parse(string[] arguments, IFileSystem fileSystem)
-        {
-            return new BaseParameters(arguments, fileSystem);
-        }
+        protected abstract void WithOptions(List<Option> options);
```
laedit: I'd prefer that this function doesn't pass the options list but only get the specific options and add them to the main list.
Or is it necessary for something that I didn't see (yet)?

src/Pretzel.Logic/Commands/ICommandParametersExtentions.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        public static IEnumerable<ExportFactory<IHaveCommandLineArgs, CommandArgumentsExtentionAttribute>> GetCommandExtentions(this ICommandParametersExtendable commandParametersExtendable)
+        {
+            var attr = commandParametersExtendable.GetType().GetCustomAttributes(typeof(CommandArgumentsAttribute), true).FirstOrDefault();
+
+            if (attr is CommandArgumentsAttribute commandArgumentAttribute && commandParametersExtendable.ArgumentExtenders != null)
+            {
+                foreach (var factory in commandParametersExtendable.ArgumentExtenders)
+                {
+                    if (factory.Metadata.CommandNames.Contains(commandArgumentAttribute.CommandName))
```
laedit: I am not a fan to rely on strings, isn't it possible to use generics instead?

src/Pretzel.Logic/Commands/ICommandParameters.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+using System.Collections.Generic;
+using System.CommandLine;
+
+namespace Pretzel.Logic.Commands
+{
+    public interface ICommandParameters
```
laedit: Could you add some comments on public api?

src/Pretzel.Logic/Commands/ICommandParametersExtendable.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+using System.Composition;
+using System.Linq;
+using Pretzel.Logic.Extensibility;
+
+namespace Pretzel.Logic.Commands
+{
+    public interface ICommandParametersExtendable
```
laedit: I am a little bit confused, what is this interface purpose?
It seems to me that this is not used.
If it is for extensibility, I don't think that everyone should touch at export factories, it should be simplified in another class or function.

src/Pretzel.Logic/Commands/ICommand.cs
```diff
@@ -0,0 +1,9 @@
+using System.Threading.Tasks;
+
+namespace Pretzel.Logic.Commands
+{
+    public interface IPretzelCommand
```
laedit: Isn't it possible to link directly the parameters through generic, like:
```cs 
ICommand<T> where T : ICommandParameters
{
    Task Execute(T parameters);
}
```

Another remark, class should have the same name than the file.

src/Pretzel.Logic/Commands/CommandArgumentsAttribute.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Pretzel.Logic.Commands
+{
+    [MetadataAttribute]
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+    public sealed class CommandArgumentsAttribute : ExportAttribute
+    {
+        public string CommandName { get; set; }
+
+        public CommandArgumentsAttribute() : base(typeof(ICommandParameters))
```
laedit: That is very confusing 🤣 
I think it is better the other way around, the command know the parameters it needs

src/Pretzel/Commands/PretzelCommandHandler.cs
```diff
@@ -0,0 +1,60 @@
...
+            {
+                new ModelBinder(CommandParameters.GetType())
+                   .UpdateInstance(CommandParameters, bindingContext);
+
+                CommandParameters.BindingCompleted();
+            }
+
+            if (CommandParameters is ISourcePathProvider pathProvider)
+            {
+                Configuration.ReadFromFile(pathProvider.Source);
```
laedit: If the command isn't an `ISourcePathProvider` the configuration is not loaded?
We should have a default source for that case.

src/Pretzel.Logic/Configuration.cs
```diff
@@ -56,16 +60,16 @@ private void EnsureDefaults()
                 _config.Add("permalink", DefaultPermalink);
             }
 
-            _defaultsConfiguration = new DefaultsConfiguration(_config);
+            Defaults = new DefaultsConfiguration(_config);
         }
 
-        [OnImportsSatisfied]
-        internal void ReadFromFile()
+        public void ReadFromFile(string path)
```
laedit: To avoid multiple reads, what do you think of making this method static and creating a `Configuration`?

src/Pretzel.Logic/Extensibility/IHaveCommandLineArgs.cs
```diff
@@ -1,10 +1,14 @@
...
 
 namespace Pretzel.Logic.Extensibility
 {
     public interface IHaveCommandLineArgs
     {
-        void UpdateOptions(OptionSet options);
-        string[] GetArguments(string command);
+        void UpdateOptions(IList<Option> options);
+
+        void BindingCompleted();
```
laedit: It seems that this is not used, or have I missing something?

src/Pretzel.Logic/Templating/Jekyll/LiquidEngine.cs
```diff
@@ -85,13 +85,13 @@ private Hash CreatePageData(PageContext pageContext)
             return x;
         }
 
-        protected override string RenderTemplate(string templateContents, PageContext pageData)
+        protected override string RenderTemplate(string content, PageContext pageData)
```
laedit: Is this renaming part of a warning fix?

src/Pretzel.Logic/Templating/JekyllEngineBase.cs
```diff
@@ -46,7 +46,7 @@ public abstract class JekyllEngineBase : ISiteEngine
 
         protected abstract string RenderTemplate(string content, PageContext pageData);
 
-        public void Process(SiteContext siteContext, bool skipFileOnError = false)
+        public void Process(SiteContext context, bool skipFileOnError = false)
```
laedit: Is this renaming part of a warning fix?

src/Pretzel.Tests/Commands/BakeBaseCommandParametersTests`1.cs
```diff
@@ -0,0 +1,21 @@
+using System;
```
laedit: Could you remove the \`1 from the filename?

src/Pretzel.Tests/Commands/ParametersTests`1.cs
```diff
@@ -0,0 +1,39 @@
+using System;
```
laedit: Could you remove the \`1 from the filename?

src/Pretzel.Tests/Commands/PetzelBaseCommandParametersTests`1.cs
```diff
@@ -0,0 +1,171 @@
+using System;
```
laedit: Could you remove the \`1 from the filename?

src/Pretzel.Tests/Extensibility/Extensions/VirtualDirectorySupportTests.cs
```diff
@@ -52,21 +50,23 @@ public void processes_only_when_vdir_argument_is_passed()
             // arrange
             var returnThis = Substitute.For<FileBase>();
             fileSystem.File.Returns(returnThis);
-            var optionSet = new OptionSet();
-            vdirSupport.UpdateOptions(optionSet);
-            optionSet.Parse(new[] { "--vDir", "something" });
```
laedit: Sorry if I have missed it, but there is no more parsing test?

src/Pretzel.Logic/Templating/Jekyll/LiquidEngine.cs
biohazard999: warning fix: base class uses `protected abstract string RenderTemplate(string content, PageContext pageData);` hence inherited classes should use the same name.

src/Pretzel.Logic/Templating/JekyllEngineBase.cs
biohazard999: warning fix: base class uses `protected abstract string RenderTemplate(string content, PageContext pageData);` hence inherited classes should use the same name.

src/Pretzel.Logic/Commands/BaseParameters.cs
biohazard999: Agree.

src/Pretzel.Logic/Commands/ICommandParametersExtendable.cs
biohazard999: I agree on this one, I will refactor the code so it's not needed any longer.

src/Pretzel.Logic/Commands/ICommandParametersExtentions.cs
biohazard999: I thought about this one, it would only be possible if we expose all the argument types as interfaces in Pretzel.Logic. When using the export factory with metadata (without creating the instance of the extender) it is impossible to know which argument to extend.

biohazard999: This will also solve the problem for plugin authors to get the correct argument instance imported (so no double binding of the command arguments is needed)

src/Pretzel/Commands/PretzelCommandHandler.cs
biohazard999: No if the argument that is used is no `ISourcePathProvider` then no configuration is read. That's only possible in 2 cases: The command does not use arguments at all. A plugin author decides to implement ICommandArgument by hand, ignoring the fact that a `PretzelBaseCommandArguments` class should be used.
The default's provided by `DefaultsConfiguration` are still there.

src/Pretzel.Logic/Configuration.cs
biohazard999: I think it is doable if `CommandCollection` exports it's configuration.

src/Pretzel.Logic/Extensibility/IHaveCommandLineArgs.cs
biohazard999: If you mean the `BindingCompleted` method you are right. It's not used in pretzel it self (cause there are only 2 argument extentions) the purpose of this methods argument validation and path corrections (relative paths to absolutes and so on)

src/Pretzel.Logic/Configuration.cs
laedit: Maybe I am missing something, but why don't just create it in program and add it to MEF?
Or is it not permitted in MEF2?

src/Pretzel.Logic/Commands/ICommandParametersExtentions.cs
laedit: Is it not possible to use classes directly in order to avoid having to create some more interfaces?

biohazard999: I don't like the idea of using the base class for exporting. I refactored out some stuff on this. Let's look if the new design is clearer now. (Need to fix some tests first, let you know if this part is ready for review)

src/Pretzel.Logic/Commands/ICommand.cs
biohazard999: I thought about `ICommand<in T> where T : ICommandArguments`. Let's see if that is something I can do 👍

src/Pretzel.Logic/Commands/CommandArgumentsAttribute.cs
biohazard999: I thought about that as well. I'll recheck why I decided against it

src/Pretzel.Logic/Commands/ICommand.cs
biohazard999: Have another look, It's not possible to export/import open generics in `System.Composition` this way, but I refactored the whole thing to be more intuitive. I think I should also do that for the ArgumentsExtentions

src/Pretzel.Logic/Commands/CommandArgumentsAttribute.cs
biohazard999: Should be much better now.

src/Pretzel.Logic/Extensibility/IHaveCommandLineArgs.cs
laedit: It was about `BindingCompleted` thanks, for the explanation.
Regarding the options list, I'd say that the plugins author should just to pass their own options, but never have the opportunity to modify the current list of options.

src/Pretzel/Commands/PretzelCommandHandler.cs
laedit: If I am not mistaken, in these cases even if there is a configuration file, it is not loaded?
That is not a thing we want, it is very confusing.

src/Pretzel.Logic/Configuration.cs
biohazard999: That is only possible if you are inside an `OnImportSatisfied` method. But we don't know which command get's executed, so we have no way of exporting it in `CommandCollection.OnImportsSatisfied`. But there is only a single call to it.
I see another benifit: If the configuration file changes we now can reload it in the `taste` command.

src/Pretzel/Commands/BakeCommand.cs
```diff
@@ -14,16 +14,20 @@ namespace Pretzel.Commands
...
+        public BakeCommandArguments(IFileSystem fileSystem) : base(fileSystem) { }
     }
 
     [Shared]
-    [CommandInfo(CommandName = BuiltInCommands.Bake, CommandDescription = "transforming content into a website")]
-    public sealed class BakeCommand : IPretzelCommand
+    [CommandInfo(
+        CommandName = BuiltInCommands.Bake,
+        CommandDescription = "transforming content into a website",
+        CommandArgumentsType = typeof(BakeCommandArguments)
```
laedit: This information can't be retrieved by the class type generic parameter?

src/Pretzel.Logic/Configuration.cs
biohazard999: And no we cant create it in program. But it's so early in the lifecycle so i think it's fine there.

src/Pretzel/Commands/BakeCommand.cs
biohazard999: Nope, we don't have that information cause no instance is created yet.

biohazard999: And MEF don't tell us the type the metadata is defined on to use reflection to get that information.

src/Pretzel.Logic/Configuration.cs
laedit: Ok, I'd prefer to do otherwise but since we are restricted by mef2 that will do.
And the benefit is great!

src/Pretzel/Commands/PretzelCommandHandler.cs
biohazard999: It's not loaded in 2 cases: For the `VersionCommand` and the `HungryCommand`. 
Both provide no extension support. All other Commands provide this. 
If I think further a bit there are commands where the configuration don't need to be read, for example the nuget ones. I think this is okay for plugin authors to use `PretzelBaseCommandArguments` for their custom commands if they want to extend the site, and go with `BaseCommandArguments` (or implement the interfaces them self) if they want to get lower.

biohazard999: But configuration it get's loaded before any command ever is created.

biohazard999: But you can't (and should not inject) `IConfiguration` into an `ICommandArgument` cause you need arguments to load the configuration. Technically you could do that in `PretzelBaseCommandArguments.BindingCompleted`, but that feels wrong cause you would break the `Liskov substitution principle`.

laedit: Ooooooooooh ok, I get it now, sorry!

src/Pretzel.Logic/Commands/CommandArgumentsAttribute.cs
laedit: It is, thanks!

src/Pretzel.Logic/Extensibility/IHaveCommandLineArgs.cs
biohazard999: now they need to provide their own collection

src/Pretzel.Tests/Extensibility/Extensions/VirtualDirectorySupportTests.cs
biohazard999: Added `AzureHostSupportArgumentsTests` and `VirtualDirectorySupportArgumentsTests` and found 1 bug 😀 and fixed it

src/Pretzel.Logic/Commands/ICommandParametersExtendable.cs
biohazard999: Not needed any longer

src/Pretzel.Logic/Commands/ICommandParametersExtentions.cs
biohazard999: Should be done.

src/Pretzel.Logic/Commands/ICommandParameters.cs
biohazard999: Done.

src/Pretzel.Logic/Commands/ICommand.cs
biohazard999: If I do that for `ArgumentExtensions` i will need to export an interface for every arguments class.

laedit: I don't think so, that seems counterproductive.

src/Pretzel/Commands/BakeCommand.cs
laedit: I am frustrated to have to repeat an info but if we don't have the choice then so be it...

src/Pretzel.Logic/Commands/ICommand.cs
biohazard999: Hmmm but now after implementing it doesn't look that bad.

src/Pretzel/Commands/BakeCommand.cs
biohazard999: I think we could if we go for non attribute based composition (by using the convention based approach) but that would be out of scope for this PR

laedit: I was thinking of that too, but haven't got the time to do some tries today.
To me, it is a part of this PR since composition is essentially used for commands and arguments so it is justified.
But I could understand if you prefer to finish this PR before starting the next big one.

biohazard999: I think we can support both models in the future, without having to make an breaking change for plugin authors. So I think it is out of scope.
Cause if we enforce some kind of convention, people just can remove parts of the attributes, so it's is an additional way (from the technical perspective we are the ones that provide the missing information).

laedit: Well since we already are breaking (nearly) everything I think it is best to also simplify everything.
If the convention allows us to easily maintain the project and the plugin author to create plugin without having to use a lot of attributes we should go for it.

biohazard999: Okay let's address this with a separate issue and put it on the v1.0 list: #353

laedit:
So sorry that it took me so long, but I needed to try some things about `System.CommandLine` to understand it and be comfortable to review your PR, you did an excellent job!
That said there are some small things that I think can be simplified (and more with the #353), I will post some comments.

src/Pretzel/Commands/CommandCollection.cs
```diff
@@ -12,56 +12,72 @@ namespace Pretzel.Commands
...
-        public void OnImportsSatisfied()
+        internal void OnImportsSatisfied()
         {
-            commandMap = new Dictionary<string, ICommand>(Commands.Length);
+            RootCommand = new RootCommand();
 
             foreach (var command in Commands)
             {
-                if (!commandMap.ContainsKey(command.Metadata.CommandName))
+                var subCommand = new Command(command.Metadata.CommandName, command.Metadata.CommandDescription);
```
laedit: I think we can simplify all this by exporting directly `Command`s: each command will inherit from `Command` and thus 
handle their options.
This way we can just `ImportMany` commands and add them to the `RootCommand`.

Regarding the arguments extensibility, we can use the interface `ICommandArgumentsExtension` with a property or method indicating the type of the command: `public Type For() => typeof(BakeCommand);`
I wanted to use a generic parameter but I haven't found a way to export it correctly.

src/Pretzel/Program.cs
```diff
@@ -1,89 +1,160 @@
...
+            }
+            catch (Exception ex)
+            {
+                Tracing.Error(ex.Message);
+                WaitForClose();
+                return -1;
+            }
+        }
+
+        internal static string[] PatchSourcePath(string[] args)
```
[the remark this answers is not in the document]
biohazard999: I'm not sure if `Arguments` can be used in this case. The reason why this exist is: in previous versions it was possible to write: `pretzel bake c:\path\to\blog` as well as `pretzel bake -s c:\path\to\blog`. If they can be optional it may be possible, but I have no idea how.

src/Pretzel/Program.cs
```diff
@@ -1,89 +1,160 @@
...
+                try
+                {
+                    InitializeTrace(debug);
+                    Tracing.Info("starting pretzel...");
+                    Tracing.Debug("V{0}", Assembly.GetExecutingAssembly().GetName().Version);
 
-                if (program.parameters.Help || !args.Any())
+                    using (var host = Compose(debug, safe, source))
+                    {
+                        var program = host.GetExport<Program>();
```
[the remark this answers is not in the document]
biohazard999: You are right. It's so missleading that `SatisfyImports` collects exports as well. I was unsure about `IFileSystem`. Fixed.

src/Pretzel/Program.cs
```diff
@@ -1,89 +1,160 @@
...
+        private async Task<int> Run(IEnumerable<Option> globalOptions, string[] args)
         {
-            Commands.WriteHelp(parameters.Options);
-            WaitForClose();
-        }
+            try
+            {
+                foreach (var option in globalOptions)
+                {
+                    CommandCollection.RootCommand.AddOption(option);
```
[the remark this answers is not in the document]
biohazard999: You are right. Fixed.

src/Pretzel/Program.cs
```diff
@@ -1,89 +1,160 @@
...
+        {
+            new Option("--debug", "Enable debugging")
+            {
+                Argument = new Argument<bool>()
+            },
+            new Option("--safe", "Disable custom plugins")
+            {
+                Argument = new Argument<bool>()
+            },
+            new Option(new[] { "--source", "-source", "-s", "--s" }, "The path to the source site (default current directory)")
```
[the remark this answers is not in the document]
biohazard999: I was following the old argument's. But the library is clever enough to cover that.

src/Pretzel/Program.cs
```diff
@@ -1,89 +1,160 @@
...
+                foreach (var option in globalOptions)
+                {
+                    CommandCollection.RootCommand.AddOption(option);
 
-        private void Run()
-        {
-            if (Commands[parameters.CommandName] == null)
+                    foreach (var command in CommandCollection.RootCommand.OfType<Command>())
+                    {
+                        command.AddOption(option);
```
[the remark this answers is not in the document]
biohazard999: This is done to avoid duplication. Those get mapped to [PretzelBaseCommandArguments](https://github.com/Code52/pretzel/pull/347/files#diff-ff4a8a22a60456b2e4a9dd66434ab628R38) or any `Argument` that provides any of the global options by convention. `Source` `Debug` and `Safe` by now.

src/Pretzel/Program.cs
```diff
@@ -1,89 +1,160 @@
...
-        private static void Main(string[] args)
-        {
-            Args = args;
-            var program = new Program();
+            args = PatchSourcePath(args);
 
-            program.InitializeTrace();
-            Tracing.Info("starting pretzel...");
-            Tracing.Debug("V{0}", Assembly.GetExecutingAssembly().GetName().Version);
+            foreach (var option in GlobalOptions)
```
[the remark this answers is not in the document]
biohazard999: I did this to support the built in help command even if a command failes. Cause it's used in the startup phase (we need the source path for plugins, and debug and safe to control it's behavior). They map to the arugments in the root command handler `CommandHandler.Create(async (bool debug, bool safe, string source)`. [Program](https://github.com/Code52/pretzel/pull/347/files#diff-40e719901a3e1c2feb8c537746f27bfcR59)

src/Pretzel/Commands/CommandCollection.cs
```diff
@@ -12,56 +12,72 @@ namespace Pretzel.Commands
...
-            {
-                ICommand command;
-                commandMap.TryGetValue(name.ToLower(System.Globalization.CultureInfo.InvariantCulture), out command);
-                return command;
-            }
+            get => argumentExtensions ?? Array.Empty<ExportFactory<ICommandArgumentsExtension, CommandArgumentsExtensionAttribute>>();
+            set => argumentExtensions = value;
         }
 
+        [Export]
```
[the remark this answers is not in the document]
biohazard999: Correct. Fixed

src/Pretzel/Commands/CommandCollection.cs
biohazard999: I did not want to do that for a reason: If we import the `commands` directly we import all dependencies with it. If a plugin provides a custom command, but is broken in some kind of way (for example it's incompatible after a new release of pretzel) it will break the whole application, even the custom command is not called by the end user.

The way it is implemented right now does instanciate as less as possible via MEF before we know what command the end user picked.

biohazard999: > Regarding the arguments extensibility, we can use the interface ICommandArgumentsExtension with a property or method indicating the type of the command: public Type For() => typeof(BakeCommand);
I wanted to use a generic parameter but I haven't found a way to export it correctly.

That is possible, but then we need to move the commands from `Petzel` to `Pretzel.Logic` and make them public. That way we can drop the `IBakeCommandArguments`. That's a reasonable change in my opinion cause we don't throw around all those argument interfaces.

src/Pretzel/Program.cs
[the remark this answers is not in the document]
laedit: I haven't succeeded on using an Argument on a RootCommand so this will do.

src/Pretzel/Program.cs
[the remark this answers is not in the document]
laedit: Since they are added to the subcommands, I don't think that they are needed in the RootCommand for the help.
And they can be used directly through their `Parse` method without the need to attach them to a command.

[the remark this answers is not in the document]
biohazard999: I'm not sure about this:

```cs
   var parseResult = new Parser(GlobalOptions).Parse(args);

            if (parseResult.Errors.Count > 0)
            {
                foreach (var error in parseResult.Errors)
                {
                    // Can' use Trace here cause we need Debug?!
                }
                return -1;
            }

            rootCommand.Handler = CommandHandler.Create(async (bool debug, bool safe, string source) =>
            {
                try
                {
                    InitializeTrace(debug);
                    Tracing.Info("starting pretzel...");
```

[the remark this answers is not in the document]
biohazard999: I could initialize Trace with debug first, and reinitialize later with the argument provided by the user. 🤷‍♂️

src/Pretzel/Commands/CommandCollection.cs
laedit: > I did not want to do that for a reason: If we import the `commands` directly we import all dependencies with it. If a plugin provides a custom command, but is broken in some kind of way (for example it's incompatible after a new release of pretzel) it will break the whole application, even the custom command is not called by the end user.

I understand your view but in the case of this small app I don't think it's worth so much complexity: if a plugin is failing the use will necessarily have to replace or update it since it was needed by the user.

> The way it is implemented right now does instanciate as less as possible via MEF before we know what command the end user picked.

I think we can delay and handle the exception through lazy imports.
 
> If we use the change you suggested, we will need to instanciate the commands or do some kind of reflection trickery, instead of using the path `MEF` provides.

Sorry but I don't understand why we will need to instanciate the commands, they will be available directly through the import.
You can see an example in my testing repo: the [custom command](https://github.com/laedit/CommandLineExtensibilityWithMEF2/blob/master/CommandLineExtensibility/BuiltinCommand2.cs) is [exported as a command](https://github.com/laedit/CommandLineExtensibilityWithMEF2/blob/master/CommandLineExtensibility/Program.cs#L104) (through convention but the same can be achieved through attributes) and is then [imported as command](https://github.com/laedit/CommandLineExtensibilityWithMEF2/blob/master/CommandLineExtensibility/Starter.cs#L60).
Or am I missing something?


> That is possible, but then we need to move the commands from `Petzel` to `Pretzel.Logic` and make them public. That way we can drop the `IBakeCommandArguments`. That's a reasonable change in my opinion cause we don't throw around all those argument interfaces and can just export the arguments directly.

Good to me.

src/Pretzel/Program.cs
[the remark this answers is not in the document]
laedit: Since debug only sets the minimal trace level to debug I don't think it is needed to trace the parse errors.
`InitializeTrace` can be separated into two parts: 
1. `Tracing.SetTrace(ConsoleTrace.Write);` as soon as possible
2. `Tracing.SetMinimalLevel(TraceLevel.Debug);` once the parsing is done

src/Pretzel/Commands/CommandCollection.cs
biohazard999: > > I did not want to do that for a reason: If we import the `commands` directly we import all dependencies with it. If a plugin provides a custom command, but is broken in some kind of way (for example it's incompatible after a new release of pretzel) it will break the whole application, even the custom command is not called by the end user.
> 
> I understand your view but in the case of this small app I don't think it's worth so much complexity: if a plugin is failing the use will necessarily have to replace or update it since it was needed by the user.

I think this from an end users perspective (I know its hypothetical for now but realistic in my opinion):

A users starts using pretzel. He want's to start his blog cause he is frustrated using his current blog software for example medium.

We have a plugin `Pretzel.Plugins.MediumImporter`.
He pulls in the dependency, imports the blog:

`pretzel import medium --foo --bar`
runs
`pretzel taste`

is happy blogging all a long, never needs the import plugin again.

We release a version, plugin stops working for whatever reason.
Suddenly `pretzel taste` will break. Enduser is puzzeled why it stopped working.

Another reason is: Almost every dependency in the whole app get's instanciated, that's wasteful.

But maybe I'm to parnoid this time 😅


> 
> > The way it is implemented right now does instanciate as less as possible via MEF before we know what command the end user picked.
> 
> I think we can delay and handle the exception through lazy imports.

That's a possibility. Still persists the problem of instanciating all the dependencies in the app. as mentioned above.

> 
> > If we use the change you suggested, we will need to instanciate the commands or do some kind of reflection trickery, instead of using the path `MEF` provides.
> 
> Sorry but I don't understand why we will need to instanciate the commands, they will be available directly through the import.
> You can see an example in my testing repo: the [custom command](https://github.com/laedit/CommandLineExtensibilityWithMEF2/blob/master/CommandLineExtensibility/BuiltinCommand2.cs) is [exported as a command](https://github.com/laedit/CommandLineExtensibilityWithMEF2/blob/master/CommandLineExtensibility/Program.cs#L104) (through convention but the same can be achieved through attributes) and is then [imported as command](https://github.com/laedit/CommandLineExtensibilityWithMEF2/blob/master/CommandLineExtensibility/Starter.cs#L60).
> Or am I missing something?

If you import something it will get instanciated. If we import lazy we have still no way to get to the arguments from the command without instanciating it. 🤷‍♂️

> 
> > That is possible, but then we need to move the commands from `Petzel` to `Pretzel.Logic` and make them public. That way we can drop the `IBakeCommandArguments`. That's a reasonable change in my opinion cause we don't throw around all those argument interfaces and can just export the arguments directly.
> 
> Good to me.

Good let's do this 👍

src/Pretzel/Program.cs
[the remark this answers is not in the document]
biohazard999: Good to me!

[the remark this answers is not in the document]
biohazard999: Better?

src/Pretzel/Program.cs
```diff
@@ -1,89 +1,166 @@
...
+            var command = new RootCommand()
+            {
+                TreatUnmatchedTokensAsErrors = false
+            };
 
-            if (parameters.Debug)
+            foreach (var option in GlobalOptions)
+                command.AddOption(option);
+
+            var parseResult = new Parser(command).Parse(args);
```
laedit: You could pass directly the options in `Parser`.

src/Pretzel/Program.cs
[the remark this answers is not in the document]
laedit: Excellent :)

src/Pretzel/Commands/CommandCollection.cs
laedit: I see your point and I admit I have never thought too much about plugin broked or instantiate all the things since I have inherited the project.

So, ok to protect the end-user on this, but my only regret in this is the complexity to relate a command and his arguments.
Is it possible to "merge" the command and his arguments?
Maybe with an interface with:
- a method returning his options
- an execute method
And keeping the CommandInfo attribute with the name and description.

src/Pretzel.Logic/Commands/CommandInfoAttribute.cs
```diff
@@ -1,14 +1,18 @@
...
 using System.Composition;
 
-namespace Pretzel.Commands
+namespace Pretzel.Logic.Commands
 {
     [MetadataAttribute]
     [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
     public sealed class CommandInfoAttribute : ExportAttribute
     {
         public string CommandName { get; set; }
```
laedit: Since we are already breaking (almost) all the things, what do you think of taking the opportunity to get rid of the redundant `Command` part?

biohazard999: I'm not sure, do you mean the namespace or the naming in general?

src/Pretzel/Program.cs
biohazard999: Nope, that's not possible, cause it will tread unrecognized options/commands as errors

src/Pretzel/Commands/CommandCollection.cs
biohazard999: I am open on this, and I tought a lot about it. I think having a clear boundary between arguments and commands is a good architecture. 
I can refactor it in this shape, but let's see how it looks like after the move to pretzel.logic. 

I still think it's a good idea to keep it appart, cause it was really hard to get it working (and the tests are tons of times easier to reason about) but you are the maintainer of the project, so you have the last word

biohazard999: Another thing is extensibility, if we keep the arguments separate, it's easier to extend. But it's more or less a design choice

biohazard999: Another thing is extensibility, if we keep the arguments separate, it's easier to extend. But it's more or less a design choice

src/Pretzel.Logic/Commands/CommandInfoAttribute.cs
laedit: The naming of properties for this attribute, like using `Name` instead of `CommandName`.
I think that `[CommandInfo(Name...` is pretty self-explanatory without having to repeat `Command`, but maybe it is just me.

src/Pretzel/Program.cs
laedit: Ah, I forgot that, sorry.

src/Pretzel.Logic/Commands/CommandInfoAttribute.cs
biohazard999: Agreed. Removed redundant naming.

laedit: Thanks!

src/Pretzel/Commands/CommandCollection.cs
biohazard999: I moved the commands to logic and removed the arguments interfaces.

laedit: Thanks, it has really helped me to get out of the fog I was regarding your proposal.
So, ok for command / arguments separation.

There is still a thing that bothers me, sorry for noticing it only now, but the command arguments extensibility have shifted his paradigm, before we were [extending a command](https://github.com/Code52/pretzel/blob/master/src/Pretzel.Logic/Extensibility/Extensions/AzureHostSupport.cs#L33) and now we are [extending a command arguments](https://github.com/biohazard999/pretzel/blob/topic/System.CommandLine.Experimental-346/src/Pretzel.Logic/Extensibility/Extensions/AzureHostSupport.cs#L14).
To me it simple more intuitive to extend a command than a command's arguments.
Is the change from a technical reason or is it from your point of view?

biohazard999: Basically it's the same story as above.

You can extend an command without the need of a specifying an additional argument.
I don't know which command is extended without instanciating the command with its extentions.

Therefore: As an plugin author: If you like to provide additional command line arguments you extend the arguments. If you extend the command/engine/whatever you import the additional arguments and extend pretzel.

But basically nothing hinders the plugin author to implement `ICommandArgumentsExtension` on the extentsion it self:

```cs
[Export]
    [Shared]
    [CommandArgumentsExtension(CommandArgumentTypes = new[] { typeof(IRecipeCommandArguments) })]
    [Export(typeof(IAdditionalIngredient))]
    public class AzureHostSupport : IAdditionalIngredient, ICommandArgumentsExtension
    {
        public IList<Option> Options { get; } = new[]
        {
            new Option(new [] { "--azure", "-azure" }, "Enables deploy to azure support")
            {
                Argument = new Argument<bool>()
            }
        };

        public void BindingCompleted()
        {
            //Not used
        }
        public bool Azure { get; set; }

        private readonly IFileSystem fileSystem;
        private readonly IAssembly assembly;

        [ImportingConstructor]
        public AzureHostSupport(IFileSystem fileSystem, IAssembly assembly)
        {
            this.fileSystem = fileSystem;
            this.assembly = assembly;
        }

        public void MixIn(string directory)
        {
            if (!Azure) return;
            // Move everything under the _source folder
            var sourceFolder = Path.Combine(directory, "_source");
            if (!fileSystem.Directory.Exists(sourceFolder))
            {
                fileSystem.Directory.CreateDirectory(sourceFolder);
            }

            foreach (var file in fileSystem.Directory.GetFiles(directory))
            {
                var trimStart = file.Replace(directory, string.Empty).TrimStart(Path.DirectorySeparatorChar);
                fileSystem.File.Move(file, Path.Combine(sourceFolder, trimStart));
            }

            foreach (var directoryToMove in fileSystem.Directory.GetDirectories(directory).Where(n => new DirectoryInfo(n).Name != "_source"))
            {
                var trimStart = directoryToMove.Replace(directory, string.Empty).TrimStart(Path.DirectorySeparatorChar);
                fileSystem.Directory.Move(directoryToMove, Path.Combine(sourceFolder, trimStart));
            }

            fileSystem.File.WriteAllText(Path.Combine(directory, @"Shim.cs"), Properties.RazorAzure.Shim);
            fileSystem.File.WriteAllText(Path.Combine(directory, @"Shim.csproj"), Properties.RazorAzure.ShimProject);
            fileSystem.File.WriteAllText(Path.Combine(directory, @"Shim.sln"), Properties.RazorAzure.ShimSolution);

            var currentPath = assembly.GetEntryAssemblyLocation();
            var destination = Path.Combine(directory, "Pretzel.exe");
            if (!fileSystem.File.Exists(destination))
            {
                fileSystem.File.Copy(currentPath, destination);
            }

            Tracing.Info("Shim project added to allow deployment to azure websites");
        }
    }
```

How a command is extensible is decided by the command it self. 

Until the point where the command is created only arguments exist. 
They provide information about possible values, descriptions, help text, argument validation etc.

Hope that makes sense.

biohazard999: Of course we could change the `CommandArgumentsExtensionAttribute` to accept the `typeof(BakeCommand)` instead of `typeof(BakeCommandArguments)` cause we have all the metadata on the command to find the correct argument to extend.

laedit: > Of course we could change the `CommandArgumentsExtensionAttribute` to accept the `typeof(BakeCommand)` instead of `typeof(BakeCommandArguments)` cause we have all the metadata on the command to find the correct argument to extend.

If that is possible without questioning the rest of your work please do it, otherwise leave as it is.
Again thanks a lot for your fabulous work! 😄

src/Pretzel.Logic/Commands/BuiltInCommands.cs
```diff
@@ -0,0 +1,13 @@
+namespace Pretzel.Logic.Commands
+{
+    public static class BuiltInCommands
+    {
+        public const string Bake = "bake";
```
laedit: Sorry, one last nitpick: I am not sure about the need for these constants, are they used internally and I have missed their usage?
Because for the extensibility, all go through types, right?

biohazard999: Those were from the first round where we used string based mapping instead of types. Not used -> deleted.

src/Pretzel/Commands/CommandCollection.cs
biohazard999: Okay, now it's possible, but i need to add the command type to the `CommandInfoAttribute` (I don't know why I didn't come up with this idea earlier). It's ugly and redundant as hell (I think that was the reason) but I hope we fix this with #353

laedit: I was thinking the same, but yes I think we can automatically export the command with a metadata including his type.
If that's not possible and if that is ok with you, we will rollback only for this attribute property and uses only the CommandArguments for extensibility.

laedit:
We have finally pulled this through!
Again, thanks a lot for your hard work and having endured me 😄

PR #349, opened 2019-09-28
Use Microsoft.CodeAnalysis.FxCopAnalyzers

biohazard999:
This just add's the `Microsoft.CodeAnalysis.FxCopAnalyzers` packages to `Pretzel` and `Pretzel.Logic`. 
It will help to avoid unnecessary allocations as well as possible `NRE`s and possible problems with formatting and globalization.


laedit:
Thanks, that's an excellent idea!
Can you add it to `Pretzel.Logic` too?
And if there are some warnings that could be easily fixed, can you fix them in this PR?

biohazard999:
> Thanks, that's an excellent idea!
> Can you add it to `Pretzel.Logic` too?
> And if there are some warnings that could be easily fixed, can you fix them in this PR?

🤦‍♂ I removed it from Tests and also from Logic. Done.

There are some related to globalization but not 100% sure how and if they are fixable.

Most are related to methods that can be made static.
I'll look into the easy ones.

biohazard999:
I'd like to wait to fix the warnings until #347 is merged cause I fixed already some warnings their to avoid conflicts.

laedit:
Works for me.

PR #351, opened 2019-09-29
Replace RazorEngine with RazorEngine.NetCore

biohazard999:
This relates to #337.

Seams like the PR to [fouadmess](https://github.com/fouadmess/RazorEngine/pull/2/files) will not be merged (no reaction).
So I decided to fork it and publish a forked [package](https://www.nuget.org/packages/RazorEngine.NetCore.Pretzel)

This also removed: `Microsoft.Windows.Compatibility` cause it wasn't needed at all.


biohazard999:
Seams like this one is easier than I thought, but there seams like a difference I can't figure out.
There is a change how the model is resolved, but I'm not confident enough what causes this.

laedit (approved):

PR #359, opened 2019-10-20
Topic/remove nowin

biohazard999:
This replaces Nowin with AspnetCore 2.2 and should be the last PR we need to go full netstandard2.0.

relates to #337


src/Pretzel/Properties/launchSettings.json
```diff
@@ -2,7 +2,7 @@
   "profiles": {
     "Pretzel": {
       "commandName": "Project",
-      "commandLineArgs": "taste -s c:\\f\\git\\Scissors.Blog --debug"
+      "commandLineArgs": "taste -s c:\\f\\git\\Scissors.Blog"
```
laedit: Sorry I haven't seen this earlier, but shouldn't the command line used a relative path in order to be used by anyone?

biohazard999: I did notice either. Either we should delete and ignore the File all together, or provide a sample Project by Default

laedit (approved):

Issue #338, opened 2019-09-17
Release as dotnet core global tool


laedit:
In addition to existing assets.

PR #360, opened 2019-10-24
Package Pretzel as a global tool

biohazard999:
Packs Pretzel as a global tool.

relates to #338

@laedit should be ready for review :)


laedit:
Seems good to me, can you just update `appveyor.yml` to add `Pretzel.Tool` in the artifacts and in the nuget deploy?
Since I think dotnet global tools need to be available from nuget, right?

biohazard999:
> Seems good to me, can you just update `appveyor.yml` to add `Pretzel.Tool` in the artifacts and in the nuget deploy?
> Since I think dotnet global tools need to be available from nuget, right?

So you mean like so? I wasn't aware that you can specify them in `appveyor.yml` as well

appveyor.yml
```diff
@@ -71,3 +74,10 @@ deploy:
     artifact: Pretzel.Logic.nupkg, Pretzel.Logic.symbols.nupkg
     on:
       appveyor_repo_tag: true
+
+  - provider: NuGet
+    api_key:
+      secure: NmM24h0TPsfINvy/1Qr4iJkmRjVztuY5umCJ6mzpxSjvuKl48AXjGEcgow+iJwL8
+    artifact: Pretzel.Tool.nupkg, Pretzel.Tool.nupkg
```
laedit: No need to upload it twice :)

biohazard999: Sorry I was really tired yesterday. It was supposed to include the symbols package.

appveyor.yml
```diff
@@ -71,3 +74,10 @@ deploy:
     artifact: Pretzel.Logic.nupkg, Pretzel.Logic.symbols.nupkg
     on:
       appveyor_repo_tag: true
+
+  - provider: NuGet
+    api_key:
+      secure: NmM24h0TPsfINvy/1Qr4iJkmRjVztuY5umCJ6mzpxSjvuKl48AXjGEcgow+iJwL8
+    artifact: Pretzel.Tool.nupkg, Pretzel.Tool.symbols.nupkg
```
laedit: `Pretzel.Tool.symbols.nupkg` haven't been declared in `appveyor.yml` and the content of `Pretzel.Tool.symbols.nupkg` and `Pretzel.Tool.nupkg` looks similar, so I don't know if it is because appveyor upload the same file for the two artifacts or if global tool doesn't handle symbols.

biohazard999: Seams really like so. Should I try to include symbols in the tools package and omit the symbols package at all? Or do we need the symbols in practice?

laedit: I think we can omit the symbols for a global tool.

biohazard999: Should be fixed.

